Issue #14, opened 2020-08-10
Crashes when no skippedIds are specified


zach-klippenstein:

pyricau:
Fixed by #16

PR #16, opened 2020-08-10
API cleanup (aka bulldozer mode)

pyricau:


radiography/src/main/java/radiography/FocusedWindowViewFilter.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import android.view.View
+
+/**
+ * Filters out root views that don't currently have the window focus from the output of [Radiography.scan].
+ */
+class FocusedWindowViewFilter : ViewFilter {
+
+  override fun matches(view: View): Boolean {
+    return view.parent != null || view.hasWindowFocus()
```
pyricau, line 11: need to double check that `view.parent != null` actually works here, we need a sample module with an android app.

also we should have UI tests ;)

radiography/api/radiography.api
```diff
@@ -7,41 +7,33 @@ public final class com/squareup/radiography/BuildConfig {
...
-	public final fun textFieldMaxLength (I)Lradiography/Xrays$Builder;
+public final class radiography/ViewFilter$All : radiography/ViewFilter {
+	public static final field INSTANCE Lradiography/ViewFilter$All;
+	public fun matches (Landroid/view/View;)Z
 }
 
-public final class radiography/Xrays$Companion {
-	public final fun create ()Lradiography/Xrays;
-	public final fun withSkippedIds ([I)Lradiography/Xrays;
+public final class radiography/ViewsKt {
```
pyricau, line 35: ViewsKt is fine right? It's an extension function that delegates to Radiography.scan so not needed for Java

radiography/src/main/java/radiography/Radiography.kt
```diff
@@ -0,0 +1,221 @@
...
+   * match specific ids (e.g. a debug drawer). Use [FocusedWindowViewFilter] to keep only the
+   * views of the currently focused window, if any.
+   */
+  @JvmStatic
+  fun scan(
+    rootView: View? = null,
+    includeTextViewText: Boolean = false,
+    textViewTextMaxLength: Int = Int.MAX_VALUE,
+    viewFilter: ViewFilter = ViewFilter.All
+  ): String = buildString {
```
pyricau, line 49: Ok so what's not there:

* Separate "go through the view hierarchy" from "here's how to render a single view" or "here's the attributes I want to collect". If we did either of these last two things, we wouldn't want includeTextViewText and textViewTextMaxLength in the API. So, solid breaking API change, so we maybe should make our mind up before release.
* Separate out "scanning a view hierarchy" from "scanning any kind of hierarchy" (e.g. Compose).  I believe this can be extracted out as a separate API  that this uses. The only question is whether "Radiography" should be for views or the more general thing and then we have a specialized version for views. Or the "specialization" is a param..

radiography/src/main/java/radiography/Radiography.kt
```diff
@@ -0,0 +1,221 @@
...
+
+    val matchingRootViews = rootViews.filter(viewFilter::matches)
+
+    val config = Config(
+        includeTextViewText, textViewTextMaxLength, viewFilter
+    )
+
+    for (view in matchingRootViews) {
+      if (length > 0) {
+        append("\n")
```
zach-klippenstein, line 68: Nit: `appendln()`

radiography/src/main/java/radiography/Radiography.kt
```diff
@@ -0,0 +1,221 @@
...
+    if (view is ViewGroup) {
+      val lastNonSkippedChildIndex = findLastNonSkippedChildIndex(view)
+      val lastChildIndex = view.childCount - 1
+      for (index in 0..lastChildIndex) {
+        if (index == lastNonSkippedChildIndex) {
+          lastChildMask = lastChildMask or (1 shl depth).toLong()
+        }
+        val child = view.getChildAt(index)
+        // Never null on the main thread, but if called from another thread all bets are off.
+        if (child != null && viewFilter.matches(child)) {
```
zach-klippenstein, line 115: Nit: `child?.let { if (viewFilter.matches(it) { …`

radiography/src/main/java/radiography/Radiography.kt
```diff
@@ -0,0 +1,221 @@
...
+    return -1
+  }
+
+  @TargetApi(CUPCAKE)
+  private fun Config.viewToString(
+    result: StringBuilder,
+    view: View
+  ) {
+    result.append("${view.javaClass.simpleName} { ")
+    if (view.id != -1 && view.resources != null) {
```
zach-klippenstein, line 139: Shouldn't this be `View.NO_ID` instead of `-1`? If not, it would still be better to extract our own private constant.

radiography/src/main/java/radiography/SkipIdsViewFilter.kt
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Arrays
+
+/**
+ * Filters out views with ids matching [skippedIds] from the output of [Radiography.scan].
+ */
+class SkipIdsViewFilter(private vararg val skippedIds: Int) : ViewFilter {
+
+  override fun matches(view: View): Boolean {
+    val viewId = view.id
+    return (viewId == View.NO_ID || Arrays.binarySearch(skippedIds, viewId) < 0)
```
zach-klippenstein, line 13: Nit: I think `binarySearch` is an extension on arrays in Kotlin (my bad for not doing this in the kotlin conversion PR)

radiography/src/main/java/radiography/Views.kt
```diff
@@ -0,0 +1,18 @@
...
+ * Extension function for [Radiography.scan] when scanning starts from a specific view.
+ * @see Radiography.scan
+ */
+fun View?.scan(
+  includeTextViewText: Boolean = false,
+  textViewTextMaxLength: Int = Int.MAX_VALUE,
+  viewFilter: ViewFilter = ViewFilter.All
+): String {
+  if (this == null) {
+    return "null"
```
zach-klippenstein, line 15: Nit: Pull return out of `if`

radiography/src/main/java/radiography/SkipIdsViewFilter.kt
zach-klippenstein, line 13: FYI this copies the bug in #14 – not in scope for this change, we can fix after this lands.

radiography/src/main/java/radiography/ViewFilter.kt
```diff
@@ -0,0 +1,17 @@
+package radiography
+
+import android.view.View
+
+/**
+ * Used to filter out views from the output of [Radiography.scan].
+ */
+interface ViewFilter {
```
zach-klippenstein, line 8: Should we include some sort of helper (maybe an operator?) to combine filters?

radiography/src/main/java/radiography/Radiography.kt
```diff
@@ -0,0 +1,221 @@
...
+import android.view.WindowManager
+import android.widget.Checkable
+import android.widget.TextView
+import radiography.Radiography.scan
+
+/**
+ * Utility class to scan through a view hierarchy and pretty print it to a [String].
+ * Call [scan] or [View.scan].
+ */
+object Radiography {
```
zach-klippenstein, line 17: Why have this object at all, vs just having a top-level function?

radiography/src/main/java/radiography/Radiography.kt
pyricau: that'd be

```
if (child?.let { viewFilter.matches(it) } == true) {
```

sure that reads better?

pyricau: oh you mean move the whole block.

radiography/api/radiography.api
zach-klippenstein: Yep. You could even make the function `@JvmSynthetic` to hide from Java callers.

radiography/src/main/java/radiography/Radiography.kt
zach-klippenstein: There's an old PR that solves 1: https://github.com/square/radiography/tree/jlabanca/diagnosis
That API seems pretty flexible, we could do something like that.

I think scanning Compose hierarchy is a pretty specific and one-off use case, but that's a whole separate discussion that is out of scope for this PR.

radiography/src/main/java/radiography/Views.kt
jrodbx: do y'all have a preference on implicit returns vs not?  i lean towards avoiding them since they're not easily readable.

radiography/src/main/java/radiography/Radiography.kt
zach-klippenstein: `child?.let(viewFilter::matches) == true`?

radiography/src/main/java/radiography/SkipIdsViewFilter.kt
pyricau: added `|| skippedIds.isEmpty()`

radiography/src/main/java/radiography/Radiography.kt
zach-klippenstein: Idk about readability, up to you.

radiography/src/main/java/radiography/ViewFilter.kt
pyricau: oh yeah totally forgot I mean to do that.

radiography/src/main/java/radiography/FocusedWindowViewFilter.kt
zach-klippenstein: Separate PR?

radiography/src/main/java/radiography/Views.kt
zach-klippenstein: In short cases like this I think expression bodies are just as readable, but I don't have a strong opinion either way. For small one-liners I prefer expression bodies.

radiography/src/main/java/radiography/ViewFilter.kt
pyricau:
```
  /**
   * Creates a new filter that combines this filter with [otherFilter]
   */
  infix fun and(otherFilter: ViewFilter): ViewFilter {
    val thisFilter = this
    return object : ViewFilter {
      override fun matches(view: View) = thisFilter.matches(view) &&
          otherFilter.matches(view)
    }
  }
```

radiography/src/main/java/radiography/SkipIdsViewFilter.kt
zach-klippenstein: I would think the `skippedIds.binarySearch` extension would handle this case.

radiography/src/main/java/radiography/Radiography.kt
pyricau: namespace polluting, ie if we do top level function then anywhere someone types `s` we'll start showing `scan` ?

zach-klippenstein: And now when anyone types `r` they'll see `Radiography`. I don't feel strongly here though, up to you.

radiography/src/main/java/radiography/FocusedWindowViewFilter.kt
pyricau: yeah. if it turns out it didn't work I'll fix it.

radiography/src/main/java/radiography/SkipIdsViewFilter.kt
pyricau:
```
public fun IntArray.binarySearch(element: Int, fromIndex: Int = 0, toIndex: Int = size): Int {
    return java.util.Arrays.binarySearch(this, fromIndex, toIndex, element)
}
```

pyricau: unless the crash was an NPE on the array, in which case that can't happen anymore?

PR #17, opened 2020-08-10
Extract tree rendering

pyricau:
This change extracts the tree rendering part of Radiography into a separate class, paving the way for supporting other trees.

TreeStringRenderer does the heavy lifting of navigating the tree and printing the hierarchy. Note that it's not a nice functional / lazy API. This change sticks to the approach in Radiography to limit temporary object creation.

### DIFF
--- a/radiography/src/main/java/radiography/TreeStringRenderer.kt
+++ b/radiography/src/main/java/radiography/TreeStringRenderer.kt
@@ -1,0 +1,89 @@
+package radiography
+
+internal abstract class TreeStringRenderer<N> {
+
+  abstract fun StringBuilder.renderNode(node: N)
+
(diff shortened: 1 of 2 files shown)


radiography/src/main/java/radiography/TreeStringRenderer.kt
```diff
@@ -0,0 +1,91 @@
+package radiography
+
+internal abstract class TreeStringRenderer<N> {
+
+  abstract fun StringBuilder.renderNode(node: N)
+
+  abstract fun N.getChildAt(
+    index: Int
+  ): N?
```
zach-klippenstein, line 9: Nit: single line

PR #22, opened 2020-08-10
Remove filtering logic from TreeStringRenderer and use collections instead.

zach-klippenstein:
Filtering logic doesn't belong in the renderer, and while this is slightly less efficient, it shouldn't be significant.

### DIFF
--- a/radiography/src/main/java/radiography/TreeStringRenderer.kt
+++ b/radiography/src/main/java/radiography/TreeStringRenderer.kt
@@ -3,12 +3,7 @@ package radiography
 internal abstract class TreeStringRenderer<N> {
 
   abstract fun StringBuilder.renderNode(node: N)
-
-  abstract fun N.getChildAt(index: Int): N?
-
-  abstract val N.childCount: Int
-
-  abstract fun N.matches(): Boolean
+  abstract val N.children: Collection<N?>
 
   fun render(
     stringBuilder: StringBuilder,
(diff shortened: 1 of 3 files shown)


radiography/src/main/java/radiography/TreeStringRenderer.kt
```diff
@@ -3,12 +3,7 @@ package radiography
 internal abstract class TreeStringRenderer<N> {
 
   abstract fun StringBuilder.renderNode(node: N)
-
-  abstract fun N.getChildAt(index: Int): N?
-
-  abstract val N.childCount: Int
-
-  abstract fun N.matches(): Boolean
+  abstract val N.children: Collection<N?>
```
pyricau, line 6: Should we change this to `Collection<N>` ? ie not nullable?

pyricau: was nullable because of getChildAt and concurrent changes. Now the impl can take care of this.

radiography/src/main/java/radiography/Views.kt
```diff
@@ -18,3 +19,8 @@ fun View?.scan(
     Radiography.scan(this, includeTextViewText, textViewTextMaxLength, viewFilter)
   }
 }
+
+internal fun ViewGroup.childrenAsList(): List<View> = object : AbstractList<View>() {
```
pyricau, line 23: Need to account for things changing under us mid way. Let's instead create a mutable list and iterate, skipping if null?

pyricau: I'd go with just one mutable list, right where we also do the filtering.

radiography/src/main/java/radiography/TreeStringRenderer.kt
zach-klippenstein: yea, that's better

radiography/src/main/java/radiography/Views.kt
zach-klippenstein: How would things change midway, unless they're being done from another thread?

zach-klippenstein: The old code didn't make a defensive copy either.

pyricau: Another thread: yep!


E.g. when rendering the view hierarchy in a default exception handler that received a background thread crash.


Prior impl was somewhat defensive by dealing with "maybe I'll get a null child"

zach-klippenstein: Ah, that makes sense. But I thought Views threw if accessed from different threads like that?

> Let's instead create a mutable list and iterate, skipping if null?

And I'm not quite clear on what you meant. `List` is a type that indicates a read-only list, not an immutable list, so the contract still allows the list contents to change out from under us. I think we need to make the type argument nullable again though, since there's no way to actually synchronize on the view's children itself or get them all atomically.

pyricau: Sorry that wasn't clear. I meant to do a best effort + safe.

So instead of this:

```kotlin
      override val View.children: Collection<View?>
        get() = if (this is ViewGroup) childrenAsList().filter(viewFilter::matches) else emptyList()
```

Try:

```
      override val View.children: Collection<View?>
        get() = if (this is ViewGroup) {
          val children = mutableListOf<View>
          for (i in 0..childCount) {
            getChildAt(i)?.let { children += it }
          }
          children
        } else emptyList()
```

zach-klippenstein:
Made obsolete by #31.

PR #23, opened 2020-08-10
Extract out state rendering

pyricau:


radiography/src/main/java/radiography/DecoratingAppendable.kt
```diff
@@ -0,0 +1,32 @@
+package radiography
+
+internal class DecoratingAppendable(
```
zach-klippenstein, line 3: "Decorating" is generic, i'd call this "CommaSeparatedAppendable" or something.

radiography/src/main/java/radiography/Radiography.kt
```diff
@@ -43,16 +35,9 @@ object Radiography {
   @JvmStatic
   fun scan(
     rootView: View? = null,
-    includeTextViewText: Boolean = false,
-    textViewTextMaxLength: Int = Int.MAX_VALUE,
+    stateRenderers: List<StateRenderer<out Any>> = ViewStateRenderers.defaults(),
```
zach-klippenstein, line 38:
```suggestion
    stateRenderers: List<StateRenderer<*>> = ViewStateRenderers.defaults(),
```

radiography/src/main/java/radiography/StateRenderer.kt
```diff
@@ -0,0 +1,36 @@
+package radiography
+
+import radiography.StateRenderer.Companion.from
+
+/**
+ * Renders extra attributes for specifics types in the output of [Radiography.scan].
+ * Call [from] to create instances with reified types:
+ * ```
+ *  val myRenderer: StateRenderer<MyView> = from { myView, append ->
```
zach-klippenstein, line 9: A more idiomatic API and way to call this would be `val myRenderer = stateRendererFor<MyView> { myView, append ->`

`from` is a little too generic for a top-level method IMO. And it _should_ be a top-level method IMO, since there's no reason to namespace it if you're going to static-import anyway.

radiography/src/main/java/radiography/StateRenderer.kt
```diff
@@ -0,0 +1,36 @@
...
+ *    append(myView.customAttributeValue)
+ *  }
+ * ```
+ */
+abstract class StateRenderer<T>(
```
zach-klippenstein, line 14: Maybe `ViewStateRenderer`?

zach-klippenstein, line 14: Why is this abstract? Just make it final, give it a `@PublishedApi internal` constructor, and make `from` or `stateRendererFor` the only way to create one.

zach-klippenstein, line 14:
```suggestion
abstract class StateRenderer<in T>(
```

pyricau: I've been thinking we could almost move this into the more generic TreeStringRenderer? Hence it works with anything that has state.

Also annoying that one of those ViewStateRenderer would be the View class ViewStateRenderer

pyricau: Had some trouble with reified + inline and private constructors, your suggestion makes sense.

radiography/src/main/java/radiography/StateRenderer.kt
pyricau: I like the suggested name!

I still think there's a UX difference with namespacing, simply in how the IDE prioritizes auto completes (ie I think going top level will give us a higher priority and I probably wouldn't want that).

I'm not 100% certain though.

radiography/src/main/java/radiography/DecoratingAppendable.kt
pyricau: I'm half tempted to drop Appendable in favor of am interface here, since I only support two of the 3 methods, and really only care about one.

zach-klippenstein: Yea, better to define a new type than throw from a standard one.

radiography/src/main/java/radiography/StateRenderer.kt
zach-klippenstein: I'm not sure about that. There was an issue for a while where imports from `object`s didn't show up at all, but I think that's fixed, or will be in 1.4.

radiography/src/main/java/radiography/ViewStateRenderers.kt
```diff
@@ -0,0 +1,91 @@
+package radiography
+
+import android.content.res.Resources.NotFoundException
+import android.view.View
+import android.widget.Checkable
+import android.widget.TextView
+
+object ViewStateRenderers {
+
+  val viewRenderer: StateRenderer<View> = StateRenderer.stateRendererFor { view ->
```
zach-klippenstein, line 10: Static import?

PR #25, opened 2020-08-11
Add CI job to build with Kotlin 1.4.0-rc.

zach-klippenstein:


radiography/src/main/java/radiography/Radiography.kt
```diff
@@ -76,15 +76,18 @@ object Radiography {
...
       }
       val layoutParams = view.layoutParams
       val title = (layoutParams as? WindowManager.LayoutParams)?.title?.toString()
           ?: view.javaClass.name
+      @Suppress("DEPRECATION")
       appendln("$title:")
 
       val startPosition = length
       try {
+        @Suppress("DEPRECATION")
```
pyricau, line 90: What's up with those? Should we use something else?

zach-klippenstein: Yes, but we can't we're not building with 1.3 anymore. They're deprecated in 1.4.

PR #26, opened 2020-08-11
Exclude sample from binary validation.

zach-klippenstein:


build.gradle.kts
```diff
@@ -37,7 +38,19 @@ buildscript {
   }
 }
 
-apply(from = rootProject.file(".buildscript/binary-validation.gradle"))
```
zach-klippenstein, line 40: Inlined this, since it's only used once and factoring it out just made it harder to read.

pyricau: 👍 I'd copy pasted this from workflow not knowing how to do what you just did

PR #29, opened 2020-08-11
Upgrade Gradle to 6.6.

zach-klippenstein:


rjrjr:
This is a suspiciously small change. Did you actually run the wrapper upgrade command, or just patch the generated file by hand?

zach-klippenstein:
I ran `./gradlew wrapper --gradle-version=6.6 --distribution-type=all`

PR #30, opened 2020-08-11
Only access the view hierarchy from the main thread

pyricau:


zach-klippenstein:
Views can technically be created and "locked" to background threads though, right? Such views would still be racy.

pyricau:
> Views can technically be created and "locked" to background threads though, right? Such views would still be racy.

Yes and no. Yes, Views aren't tied to the main thread, they can be used from a separate single thread. However, you can't mix and match in a single hierarchy, so all the views we retrieve from the window manager are main thread only.

It's possible to create a view hierarchy in a background thread and use that for other purposes, and then call `view.scan()` in which case we'd do the wrong thing.

But we can work around that! We can call view.getHandler() and compare the looper. I'll do that.

PR #31, opened 2020-08-11
Refactor TreeStringRenderer to allow heterogenous children.

zach-klippenstein:
Replaces `TreeStringRenderer` with a top-level `renderTreeString` method and a `TreeRenderingVisitor`. Visitors implement a single `visitNode` method, which can write the current node's description to a `StringBuilder` and build a list of children nodes to recurse into.

A node's children can be of any type, and if the type is different than the current node, it just needs to pass in an appropriate visitor for that type. This will allow us to recurse into trees from Compose functions later.


zach-klippenstein:
Added unit tests.

radiography/src/main/java/radiography/ViewTreeRenderingVisitor.kt
```diff
@@ -0,0 +1,43 @@
...
+  override fun RenderingScope.visitNode(node: View) {
+    description.viewToString(node)
+
+    if (node !is ViewGroup) return
+
+    // Capture this value, since it might change while we're iterating.
+    val childCount = node.childCount
+    for (index in 0 until childCount) {
+      // Child may be null, if children were removed by another thread after we captured the child
+      // count. getChildAt returns null for invalid indices, it doesn't throw.
```
pyricau, line 26: Can be simplified once https://github.com/square/radiography/pull/30 is merged (single thread)

radiography/src/main/java/radiography/RenderTreeString.kt
```diff
@@ -0,0 +1,117 @@
...
+    val description: StringBuilder,
+    private val children: MutableList<Pair<Any?, TreeRenderingVisitor<Any?>>>
+  ) {
+
+    /**
+     * Recursively visits a child of the current node using [visitor].
+     */
+    fun <C> visitChild(
+      childNode: C,
+      visitor: TreeRenderingVisitor<C>
```
pyricau, line 47: Is this to support compose?

radiography/src/main/java/radiography/ViewTreeRenderingVisitor.kt
```diff
@@ -0,0 +1,43 @@
...
+      val child = node.getChildAt(index) ?: continue
+      if (viewFilter.matches(child)) {
+        visitChild(child)
```
pyricau, line 29: When I read `visitChild` I think "oh it's going down recursion right now" when in fact it's just collecting and will do so right after this method returns. Is it me having an incorrect expectation, or should  this be named differently?

zach-klippenstein: The laziness is an implementation detail, didn't think it mattered in the name too much. But if the name is confusing, let's change it.

radiography/src/main/java/radiography/RenderTreeString.kt
zach-klippenstein: > This will allow us to recurse into trees from Compose functions later.

Yes.

radiography/src/main/java/radiography/ViewTreeRenderingVisitor.kt
zach-klippenstein: `addChildToVisit`

PR #32, opened 2020-08-12
Add UI tests

pyricau:


radiography/src/androidTest/java/radiography/test/RadiographyUiTest.kt
```diff
@@ -0,0 +1,130 @@
...
+
+    assertThat(hierarchy).contains("window-focus:true")
+    assertThat(hierarchy).contains("Dialog title")
+
+    assertThat(hierarchy.countSubstring("window-focus")).isEqualTo(1)
+  }
+
+  private fun String.countSubstring(substring: String) = windowed(substring.length)
+      .filter { it == substring }
+      .count()
```
pyricau, line 79: 🧙‍♀️

radiography/src/androidTest/java/radiography/test/RadiographyUiTest.kt
```diff
@@ -0,0 +1,130 @@
...
+      dialog = block()
+      dialog.show()
+    }
+    dialog.waitForFocus()
+  }
+
+  /**
+   * Waits for the activity to lose focus and the dialog to gain focus.
+   */
+  private fun Dialog.waitForFocus() {
```
pyricau, line 93: 🧙‍♀️🧙‍♀️🧙‍♀️🧙‍♀️

zach-klippenstein: Should these use `IdlingResource`s?

pyricau: I don't trust idling resources.. if espresso has a solid one liner for this I'll take it, but otherwise this is bare a droid and works without extra layers

PR #33, opened 2020-08-12
WIP Compose support.

zach-klippenstein:
Proof-of-concept at this point, probably can't land until Kotlin 1.4.


sample-compose/src/main/java/com/squareup/radiography/sample/compose/ComposeSampleApp.kt
```diff
@@ -0,0 +1,127 @@
...
+        Radiography.scan()
+      },
+      "Focused window" to {
+        Radiography.scan(viewFilter = FocusedWindowViewFilter)
+      },
+//      "Start from R.id.main" to {
+//        findViewById<View>(R.id.main).scan()
+//      },
+//      "Skip R.id.show_dialog" to {
+//        Radiography.scan(viewFilter = SkipIdsViewFilter(id.show_dialog))
```
zach-klippenstein, line 73: I think maybe we can use `testTag` modifiers for this purpose?

radiography/api/radiography.api
```diff
@@ -2,8 +2,6 @@ public final class com/squareup/radiography/BuildConfig {
 	public static final field BUILD_TYPE Ljava/lang/String;
 	public static final field DEBUG Z
 	public static final field LIBRARY_PACKAGE_NAME Ljava/lang/String;
-	public static final field VERSION_CODE I
-	public static final field VERSION_NAME Ljava/lang/String;
 	public fun <init> ()V
```
pyricau, line 5: Can we remove BuildConfig from the public api (not here, but in the lib)

zach-klippenstein: #36

sample-compose/src/main/java/com/squareup/radiography/sample/compose/ComposeSampleApp.kt
zach-klippenstein: Done.

zach-klippenstein:
Just rewrote all the analysis code, polished up the API and documentation, and added tests. This is ready for review.

compose-tests/README.md
```diff
@@ -0,0 +1,5 @@
+# compose-tests
+
+Contains the UI tests for Compose support. These can't live in the main radiography module since
+they require the Compose compiler to be turned on. No source non-test source code should be placed
```
pyricau, line 4: > No source non-test source code

One too many sources?

radiography/build.gradle.kts
```diff
@@ -32,7 +32,7 @@ android {
   }
 
   defaultConfig {
-    minSdkVersion(17)
+    minSdkVersion(21)
```
pyricau, line 35: This gets exported in our AAR, right? Bumping it reduces the uses for radiography. Is there a good reason here?

radiography/src/main/java/radiography/compose/ComposeLayoutRenderers.kt
```diff
@@ -0,0 +1,132 @@
...
+import androidx.compose.ui.semantics.SemanticsProperties.Text
+import androidx.compose.ui.semantics.SemanticsPropertyKey
+import radiography.AttributeAppendable
+import radiography.TypedViewStateRenderer
+import radiography.ViewStateRenderer
+import radiography.ellipsize
+
+// Note that this class can't use viewStateRendererFor, since that function is defined in the
+// ViewStateRenderers object, whose class initializer may initialize _this_ class, which will cause
+// NoClassDefFoundExceptions. This can happen when a debugger is attached.
```
pyricau, line 16: what fresh hell is this? Are you saying that if ComposeLayoutRenderers uses a function defined in ViewStateRenderers, then when loading ViewStateRenderers the classloader may also also load ComposeLayoutRenderers? If yes, something's wrong with the compiler..

pyricau: btw, why would initializing this class cause NoClassDefFoundExceptions ? Looks like everything here has `if (!isComposeAvailable) NoRenderer`

radiography/src/main/java/radiography/compose/ComposeLayoutRenderers.kt
```diff
@@ -0,0 +1,132 @@
...
+          rendered.semanticsConfiguration.forEach { (key, value) ->
+            when (key) {
+              SemanticsProperties.TestTag -> append("test-tag:\"$value\"")
+              SemanticsProperties.AccessibilityLabel -> append("label:\"$value\"")
+              SemanticsProperties.AccessibilityValue -> append("value:\"$value\"")
+              SemanticsProperties.Disabled -> append("DISABLED")
+              SemanticsProperties.Focused -> if (value == true) append("FOCUSED")
+              SemanticsProperties.Hidden -> append("HIDDEN")
+              SemanticsProperties.IsDialog -> append("DIALOG")
+              SemanticsProperties.IsPopup -> append("POPUP")
```
pyricau, line 58: why all the uppercase yelling :) ?

radiography/src/main/java/radiography/compose/SlotTables.kt
```diff
@@ -0,0 +1,31 @@
...
+import radiography.ViewTreeRenderingVisitor
+import radiography.renderTreeString
+
+/**
+ * Scans a particular [SlotTable], ie. an entire Compose tree. You probably want to use
+ * [Radiography.scan][radiography.Radiography.scan] instead.
+ */
+@ExperimentalRadiographyComposeApi
+@JvmSynthetic
+fun SlotTable.scan(
```
pyricau, line 20: what's a use case for using this public API directly? Since we're saying "you probably want scan instead", would be interesting to also mention why you'd want this.

PR #37, opened 2020-08-12
Change ViewFilter input from View to Any.

zach-klippenstein:
This allows us to write filters that match whatever the node type is
for composables as well, without having a separate filter type.

### DIFF
--- a/radiography/src/main/java/radiography/FocusedWindowViewFilter.kt
+++ b/radiography/src/main/java/radiography/FocusedWindowViewFilter.kt
@@ -7,7 +7,7 @@ import android.view.View
  */
 object FocusedWindowViewFilter : ViewFilter {
 
-  override fun matches(view: View): Boolean {
-    return view.parent?.parent != null || view.hasWindowFocus()
+  override fun matches(view: Any): Boolean {
+    return view is View && (view.parent?.parent != null || view.hasWindowFocus())
   }
 }
--- a/radiography/src/main/java/radiography/ViewFilter.kt
+++ b/radiography/src/main/java/radiography/ViewFilter.kt
@@ -1,7 +1,5 @@
 package radiography
 
-import android.view.View
-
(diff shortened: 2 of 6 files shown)


radiography/src/main/java/radiography/ViewFilter.kt
```diff
@@ -1,18 +1,16 @@
...
-
 /**
  * Used to filter out views from the output of [Radiography.scan].
  */
 interface ViewFilter {
   /**
    * @return true to keep the view in the output of [Radiography.scan], false to filter it out.
    */
-  fun matches(view: View): Boolean
+  fun matches(node: Any): Boolean
```
pyricau, line 10: sort of a cognitive mitmatch that we have a "ViewFilter" but the param is named node. Either keep the name "view" or rename the filter interface? Not sure "node" is very clear though.

zach-klippenstein: Oops, leftover from more invasive rename. I'll revert back to `view`.

radiography/src/main/java/radiography/FocusedWindowViewFilter.kt
```diff
@@ -7,7 +7,7 @@ import android.view.View
  */
 object FocusedWindowViewFilter : ViewFilter {
 
-  override fun matches(view: View): Boolean {
-    return view.parent?.parent != null || view.hasWindowFocus()
+  override fun matches(view: Any): Boolean {
```
jrodbx, line 10: should we use a sealed class instead?

a la:
```
sealed class View {
  class AndroidView: View()
  class Composable: View()
}
```

zach-klippenstein: It's not clear what `Composable` would even be yet, so we couldn't write that code at this time.

PR #38, opened 2020-08-12
Polish the public API

zach-klippenstein:
* Make `ViewFilter.and` an extension function.
* Make `ViewStateRenderers` properties static and fix case.
* Generate JVM overloads for methods with default args.
* Rename StateRenderer to `ViewStateRenderer`.


README.md
```diff
@@ -28,7 +28,7 @@ val prettyHierarchy = Radiography.scan(stateRenderers = defaultsIncludingPii)
 
 // Append custom attribute rendering
 val prettyHierarchy = Radiography.scan(stateRenderers = defaultsNoPii +
```
pyricau, line 30: missed some changes in the readme

PR #41, opened 2020-08-14
Support rendering trees nested deeper than 64 levels.

zach-klippenstein:
This is required for Compose support, which can have trees that go far beyond
64 levels deep.

image

### DIFF
--- a/radiography/src/main/java/radiography/RenderTreeString.kt
+++ b/radiography/src/main/java/radiography/RenderTreeString.kt
@@ -1,6 +1,7 @@
 package radiography
 
 import radiography.TreeRenderingVisitor.RenderingScope
+import java.util.BitSet
 
 /**
  * Renders [rootNode] as a [String] by passing it to [visitor]'s [TreeRenderingVisitor.visitNode]
@@ -10,7 +11,7 @@ internal fun <N> StringBuilder.renderTreeString(
   rootNode: N,
   visitor: TreeRenderingVisitor<N>
 ) {
-  renderRecursively(rootNode, visitor, depth = 0, lastChildMask = 0)
+  renderRecursively(rootNode, visitor, depth = 0, lastChildMask = BitSet())
 }
 
 internal abstract class TreeRenderingVisitor<in N> {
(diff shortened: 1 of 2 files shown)


radiography/src/main/java/radiography/RenderTreeString.kt
```diff
@@ -75,7 +76,7 @@ private fun <N> StringBuilder.renderRecursively(
   var newLastChildMask = lastChildMask
   for (index in 0..lastChildIndex) {
     if (index == lastChildIndex) {
-      newLastChildMask = newLastChildMask or (1 shl depth).toLong()
+      newLastChildMask = (newLastChildMask.clone() as BitSet).apply { set(depth) }
```
zach-klippenstein, line 79: This is pretty gross. Got a better idea?

radiography/src/main/java/radiography/RenderTreeString.kt
```diff
@@ -96,7 +97,7 @@ private fun StringBuilder.appendLinePrefix(
     if (parentDepth > 0) {
       append(' ')
     }
-    val lastChild = lastChildMask and (1 shl parentDepth).toLong() != 0L
+    val lastChild = lastChildMask[parentDepth]
```
zach-klippenstein, line 100: This is pretty nice though, on the other hand.

radiography/src/test/java/radiography/RenderTreeStringTest.kt
```diff
@@ -114,11 +114,11 @@ class RenderTreeStringTest {
     )
   }
 
-  @Test fun `renderTreeString handles over maximum depth tree`() {
+  @Test fun `renderTreeString handles over 64-deep tree`() {
     val tree = generateSequence(Node("leaf")) { child ->
       Node("node", Node("leaf"), child)
     }
-        .take(34)
+        .take(300)
```
zach-klippenstein, line 121: Ought to be enough for anybody…

radiography/src/main/java/radiography/RenderTreeString.kt
pyricau: Since this isn't exposed, seems ok to use a mutable bitset. We just need to set and then unset.

radiography/src/main/java/radiography/RenderTreeString.kt
pyricau: 😳 wow that's cool.

radiography/src/main/java/radiography/RenderTreeString.kt
zach-klippenstein: Derp, that's much better.

zach-klippenstein: Done.

PR #43, opened 2020-08-14
Fix ViewFilters' behavior for non-View nodes.

zach-klippenstein:
They should "match" non-`View` nodes, and only filter when the node is actually a `View`. This behavior is never exercised right now, but would cause any of the existing filters to filter out all compose nodes when compose support is added.

### DIFF
--- a/radiography/src/main/java/radiography/FocusedWindowViewFilter.kt
+++ b/radiography/src/main/java/radiography/FocusedWindowViewFilter.kt
@@ -8,6 +8,6 @@ import android.view.View
 object FocusedWindowViewFilter : ViewFilter {
 
   override fun matches(view: Any): Boolean {
-    return view is View && (view.parent?.parent != null || view.hasWindowFocus())
+    return view !is View || (view.parent?.parent != null || view.hasWindowFocus())
   }
 }
--- a/radiography/src/main/java/radiography/SkipIdsViewFilter.kt
+++ b/radiography/src/main/java/radiography/SkipIdsViewFilter.kt
@@ -8,7 +8,7 @@ import android.view.View
 class SkipIdsViewFilter(private vararg val skippedIds: Int) : ViewFilter {
 
   override fun matches(view: Any): Boolean {
-    if (view !is View) return false
+    if (view !is View) return true
     val viewId = view.id
     return (viewId == View.NO_ID || skippedIds.isEmpty() || skippedIds.binarySearch(viewId) < 0)
   }
--- a/radiography/src/main/java/radiography/ViewFilter.kt
+++ b/radiography/src/main/java/radiography/ViewFilter.kt
@@ -21,7 +21,6 @@ interface ViewFilter {
 infix fun ViewFilter.and(otherFilter: ViewFilter): ViewFilter {
   val thisFilter = this
   return object : ViewFilter {
-    override fun matches(view: Any) = thisFilter.matches(view) &&
-        otherFilter.matches(view)
+    override fun matches(view: Any) = thisFilter.matches(view) && otherFilter.matches(view)
   }
 }


zach-klippenstein:
I'm thinking it might make sense to extract a typed view filter type which does this logic, but then `ViewFilter` and `ViewStateRenderer` have very similar concerns – both have a type parameter, and both need to only be applied to instances of that type parameter – and it might make sense for both to take the same approach. On the other hand, `ViewStateRenderer`s are composed as lists, whereas `ViewFilter`s are composed by wrapping, so maybe they're different enough that it doesn't make sense to standardize on a pattern.

The two API styles that we have right now are basically:

1. An abstract, un-parameterized supertype that accepts `Any?`s and then we provide a convenience subtype or builder function that does filtering and casting. This approach is more flexible, since it allows matching of potentially multiple types, it allows implementations to be defined as classes instead of just functions (might be useful for injection), and the interface could be a fun interface in Kotlin 1.4. The subclassing allows us to create internal implementations that individual `TreeRenderingVisitor`s can be aware of and access directly by `filterIsInstancing` a list. However, it has a slightly larger API surface.

2. A final, parameterized, internal-constructor type that has first-class knowledge of its type (i.e. has a `Class` property) and a body function and we provide a builder function to create instances of it. This has a smaller API surface, but is more limited. `TreeRenderingVisitor`s that want to find specific types would need to do so based only on the `Class` property. 

I'm leaning towards 1, since it's more flexible, the "type filtering" behavior feels like an implementation detail of a particular renderer/filter, and it's more DI-friendly. Also, because this approach only requires the supertype to have a single method, it can be a `fun interface` in Kotlin 1.4 which feels convenient. I think the contracts here are clear enough, and focused enough, that we don't really need to be super controlling about how these objects are created, like 2 is.

## Example of 1

```kotlin
// API
fun interface Foo {
  fun applyTo(instance: Any?)
}
abstract class TypedFoo<T>(fooType: Class<T>) : Foo {
  abstract fun applyToTyped(instance: T)
  final override fun applyTo(instance: Any?) = …
}
inline fun <reified T> fooFor(body: (T) -> Unit): Foo = …

// Usage
class MyFoo @Inject constructor(…) : TypedFoo(TextView::class) {
  override fun applyToTyped(instance: TextView) = …
}
val myFoo: Foo = Foo { it: Any? -> … }
val myFoo: Foo = fooFor<TextView> { it: TextView -> … }
```

## Example of 2

```kotlin
// API
final class Foo<T> internal constructor(…) { … }
inline fun <reified T> fooFor(body: (T) -> Unit): Foo<T> = …

// Usage
class MyFooProvider @Inject constructor(…) {
  val foo: Foo<TextView> = fooFor<TextView> { … }
}
val myFoo: Foo<Any?> = fooFor<Any?> { it: Any? -> … }
val myFoo: Foo<TextView> = fooFor<TextView> { it: TextView -> … }
```

radiography/src/main/java/radiography/FocusedWindowViewFilter.kt
```diff
@@ -8,6 +8,6 @@ import android.view.View
 object FocusedWindowViewFilter : ViewFilter {
 
   override fun matches(view: Any): Boolean {
-    return view is View && (view.parent?.parent != null || view.hasWindowFocus())
+    return view !is View || (view.parent?.parent != null || view.hasWindowFocus())
```
pyricau, line 11: I'm curious why this over the && ?

zach-klippenstein: If a thing is not a `View`, this filter should ignore it. We need to return `true` in that case, so that other filters can still process the value.

zach-klippenstein: This should be more clear once we decide on an answer to the above comment.

zach-klippenstein:
Chatted about https://github.com/square/radiography/pull/43#issuecomment-674167587 offline. We decided to go to with **1**, so I will follow up with a separate PR to clean that up. This PR is mostly unrelated, so I'm going to merge.

PR #44, opened 2020-08-14
Refactor ViewFIlter and ViewStateRender APIs to be more consistent.

zach-klippenstein:
This makes both APIs more consistent with each other and with the proposal discussed in https://github.com/square/radiography/pull/43#issuecomment-674167587.

Both APIs now consist of an interface that doesn't have a type parameter, a `@PublishedApi internal` builder function that returns a filter/renderer given a `Class` to specialize on and a function, and an inline reified overload that doesn't take a class.

The "batteries included" instances of each type are now both included as `@JvmStatic` fields/functions in an `object` named with the plural of the interface name (e.g. `ViewFilters.skipIdsViewFilter()` returns a `ViewFilter`). These `object`s also include helper functions, such as `ViewFilter.and` and the `@PublishedApi internal` builder functions, for a clean API for Java consumers.

This PR is two commits: the first commit refactors `ViewStateRenderer`, the second does `ViewFilter`.


radiography/src/main/java/radiography/ViewStateRenderer.kt
```diff
@@ -9,23 +11,33 @@ package radiography
  *  }
  * ```
  */
-class ViewStateRenderer<in T> @PublishedApi internal constructor(
-  private val renderedClass: Class<T>,
-  private val renderer: AttributeAppendable.(T) -> Unit
+abstract class ViewStateRenderer<in T> internal constructor(
```
zach-klippenstein, line 14: @pyricau I kept this constructor internal, but I'm not sure why this class is so locked down. Why not make this an interface and allow direct implementations? I wanted to be able to override it to make a custom implementation for compose support that I could specifically search for by type, but I could also just make the renderedType property internal and search by that if you really want this to stay final.

zach-klippenstein: More complete thoughts in this comment: https://github.com/square/radiography/pull/43#issuecomment-674167587

radiography/src/main/java/radiography/ViewStateRenderer.kt
```diff
@@ -9,23 +11,33 @@ package radiography
...
+  internal val renderedClass: Class<@UnsafeVariance T>
 ) {
+  internal abstract fun AttributeAppendable.render(rendered: T)
```
zach-klippenstein, line 17: Needs to be protected (here and the ctor)

zach-klippenstein:
Chatted about https://github.com/square/radiography/pull/43#issuecomment-674167587 offline. We decided to go to with **1**, so I will probably update this PR to take that approach in both cases. Marking as draft until that's done.

radiography/src/main/java/radiography/ViewFilter.kt
```diff
@@ -9,18 +10,28 @@ public interface ViewFilter {
    * @return true to keep the view in the output of [Radiography.scan], false to filter it out.
    */
   public fun matches(view: Any): Boolean
+}
+
+/**
+ * Base class for implementations of [ViewFilter] that only want to filter instances of a specific
+ * type. Instances of other types are always "matched" by this filter.
+ */
+internal abstract class TypedViewFilter<in T : Any>(
```
zach-klippenstein, line 19: Made both these abstract type-specific classes `internal` for now. If it turns out we actually need them to be more open, we can always make this public in a minor release.

radiography/src/main/java/radiography/Radiography.kt
```diff
@@ -32,16 +32,16 @@ public object Radiography {
...
+   * to keep only the views of the currently focused window, if any.
    */
   @JvmStatic
   @JvmOverloads
   public fun scan(
     rootView: View? = null,
-    viewStateRenderers: List<ViewStateRenderer<*>> = DefaultsNoPii,
-    viewFilter: ViewFilter = ViewFilter.All
+    viewStateRenderers: List<ViewStateRenderer> = DefaultsNoPii,
+    viewFilter: ViewFilter = ViewFilters.All
```
pyricau, line 44: I just realized that `ViewFilters.All` might be interpreted as "all filters" rather than "keep all views". Maybe rename to `ViewFilters.NoFilter` ?

zach-klippenstein: Good catch, will do.

radiography/src/main/java/radiography/ViewFilters.kt
```diff
@@ -0,0 +1,50 @@
...
+    return object : ViewFilter {
+      override fun matches(view: Any) = thisFilter.matches(view) && otherFilter.matches(view)
+    }
+  }
+
+  /**
+   * Returns a [ViewFilter] that matches any instances of [T] for which [predicate] returns true.
+   */
+  @JvmStatic
+  @PublishedApi internal fun <T : Any> viewFilterFor(
```
pyricau, line 44: It looks like this s defined both here and in ViewFilter.java? Did you mean to keep this one?

radiography/src/main/java/radiography/ViewFilters.kt
```diff
@@ -0,0 +1,50 @@
...
+  @JvmField
+  public val FocusedWindowViewFilter: ViewFilter = viewFilterFor<View> { view ->
+    view.parent?.parent != null || view.hasWindowFocus()
+  }
+
+  /**
+   * Filters out views with ids matching [skippedIds] from the output of [Radiography.scan].
+   */
+  @JvmStatic
+  public fun skipIdsViewFilter(vararg skippedIds: Int): ViewFilter = viewFilterFor<View> { view ->
```
pyricau, line 24: Why rename this from `SkipIdsViewFilter` to `skipIdsViewFilter`?

radiography/src/main/java/radiography/ViewStateRenderers.kt
```diff
@@ -94,4 +94,17 @@ public object ViewStateRenderers {
       }
     }
   }
+
+  /**
+   * Creates a [ViewStateRenderer] that renders views of type [T].
+   */
+  @JvmStatic
+  @PublishedApi internal fun <T : Any> viewStateRendererFor(
```
pyricau, line 102: not public?

radiography/src/main/java/radiography/ViewFilters.kt
zach-klippenstein: The inline one in `ViewFilter.kt` calls this one. This kills two birds with one stone:

1. The inline one isn't visible to Java callers, but this one is (and vice versa, since this is internal).
2. If the inline function used the `object : TypedViewFilter` syntax directly, the anonymous object declaration would be inlined at every call site and it would create a new concrete class at each one. Delegating to another function means the same anonymous class is used for all inline call sites.

radiography/src/main/java/radiography/ViewFilters.kt
zach-klippenstein: Because it's a function now, and the style guidelines dictate it must be normal camel case. The functions in `ViewStateRenderers` follow the same pattern as well.

radiography/src/main/java/radiography/ViewStateRenderers.kt
zach-klippenstein: Nope, see reply to your comment above.

zach-klippenstein: If we actually need to call this one from Kotlin at some point, we can always change the visibility to public in a minor release.

radiography/src/main/java/radiography/ViewFilters.kt
```diff
@@ -0,0 +1,50 @@
+package radiography
+
+import android.view.View
+
+public object ViewFilters {
+
+  /** A [ViewFilter] that matches everything. */
+  @JvmStatic
+  public val All: ViewFilter = viewFilterFor<Any> { true }
```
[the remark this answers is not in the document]
zach-klippenstein: Unfortunately not, because `Nothing` is not allowed to be passed for a reified type parameter. But you're right, this isn't as efficient as it could be. I'll make this an anonymous object instead.

radiography/src/main/java/radiography/ViewFilters.kt
zach-klippenstein: Added (non-kdoc) comments for both these cases to point out the difference.

PR #45, opened 2020-08-14
Upgrade to Kotlin 1.4.

zach-klippenstein:


zach-klippenstein:
Yes, confirmed last week.

PR #47, opened 2020-08-18
Make SAM interfaces fun interfaces so they can be given as lambdas in Kotlin 1.4

zach-klippenstein:


zach-klippenstein:
ktlint's not happy. Fix is on the way, see https://github.com/pinterest/ktlint/issues/830

zach-klippenstein:
As of #51, blocked by Kotlin 1.4 upgrade again.

zach-klippenstein:
Added prefix commit to force the ktlint plugin to use ktlint version 0.38.1 and workaround #67.

PR #48, opened 2020-08-18
Add trailing commas to last items in default renderer lists.

zach-klippenstein:
This makes it easier to reorder items, and gives a cleaner diff when items are added or removed in the future.


zach-klippenstein:
ktlint's not happy. Fix is on the way, see pinterest/ktlint#830

zach-klippenstein:
As of #51, blocked by Kotlin 1.4 upgrade again.

PR #50, opened 2020-08-20
Restrict stdlib usage to 1.3.

zach-klippenstein:
This reverts the changes to stdlib call sites made in 0a0f4d266537f8bdc78fa94cc4f9166a790d2376.

This change ensures the library is still consumable from projects built with kotlin 1.3.

Some [discussion on Kotlin Slack](https://kotlinlang.slack.com/archives/C0922A726/p1597943810029800?thread_ts=1597937325.024100&cid=C0922A726) about this, using Kt 1.4 with `apiVersion = "1.3"` isn't enough, and using Kt 1.3 setting `languageLevel = "1.4"` won't work either.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -18,7 +18,8 @@ import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 import org.jlleitschuh.gradle.ktlint.KtlintExtension
 import org.jlleitschuh.gradle.ktlint.reporter.ReporterType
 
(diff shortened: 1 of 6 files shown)


zach-klippenstein:
All this config is a lot more configuration to maintain than just… not upgrading to 1.4 yet. I think I will close this and just revert #45 instead.

zach-klippenstein:
Superseded by #51.

zach-klippenstein:
Re-opening since #51 breaks the Compose sample.

PR #51, opened 2020-08-20
Revert kotlin 1.4 upgrade since it's easier than trying to explicitly support 1.3 consumers

zach-klippenstein:
See [this slack thread](https://kotlinlang.slack.com/archives/C0922A726/p1597943810029800?thread_ts=1597937325.024100&cid=C0922A726).

Replaces #50.


zach-klippenstein:
Sorry for the flip-flop, probably favoring #50 now, since it lets us build a sample app that uses Compose.

PR #60, opened 2020-08-23
Add comments explaining some of the composition analysis logic.

zach-klippenstein:


radiography/src/main/java/radiography/compose/ComposeViews.kt
```diff
@@ -102,20 +104,37 @@ private fun visitComposeView(
...
+  // asTree is provided by the Compose Tooling library. It "reads" the slot table and parses it 
+  // into a tree of Group objects. This means we're technically traversing the composable tree
+  // twice, so why not just read the slot table directly? As opaque as the Group API is, the actual
+  // slot table API is quite complicated, and the actual format of the slot table (effectively an
+  // array that stores a flattened version of a composition tree) is super low level. It's likely to
+  // change a lot between compose versions, and keeping up with that with every two-week dev release
+  // would be a lot of work. Additionally, a lot of the objects stored in the slot table are not
+  // public (eg LayoutNode), so we'd need to use even more (brittle) reflection to do that parsing.
+  // That said, once Compose is more stable, it might be worth it to read the slot table directly,
+  // since then we could drop the requirement for the Tooling library to be on the classpath.
```
pyricau, line 126: >  drop the requirement for the Tooling library to be on the classpath

Is that tooling library ok to ship in production? The view side of Radiography is meant to work in release builds as well, for crashes.

zach-klippenstein: It should be fine, as much as _anything_ from Compose is ok to ship in production right now. Most of it is related to previews, but apps that are using R8/proguard will strip all that stuff out. We're only using one method from it.

We might want to revisit this when Compose is stable, but doesn't make any sense before then.

PR #63, opened 2020-08-23
WIP Add support for only rendering certain sub-trees.

### LINKED ISSUE #58: Drop SlotTable.scan, add composable selection support
Another solution to this problem, which would be much more useful for devs, would be the ability to start the scan at some point further down in the tree, similar to passing a non-root View to scan(). However that presents a few (solvable) issues:

1. Composables don't have any implicit identity, so we'd probably need to do something like for filtering and use test tags or layout IDs to identify the root.
2. The visitor infra doesn't currently support filtering nodes above, only below, so that would have to be implemented. This could also make it possible to visit Groups directly instead of using the intermediate LayoutInfo type - I'm not sure we _should_, since that type lets us decouple a bit from compose APIs that might change a lot, but we'd have the option. 

Having thought about this more, that API makes much more sense than this one. I'll take a stab at implementing it, but would like to not block this PR on it since this API is marked as unstable anyway.

_Originally posted by @zach-klippenstein in https://github.com/square/radiography/pull/33#discussion_r474794986_

zach-klippenstein:
This change allows filters to be written which not only exclude certain subtrees, but also which _only_ include certain subtrees. 

It has three parts:
1. Change `ViewFilter`'s matches method to return a three-value type instead of a boolean, where the third value means don't include this view, but include its children.
2. Give `ComposeLayoutInfo` a pointer to its parent. 
3. Write a filter for compose that returns "include only children" for all nodes that don't have a parent which matches a certain test tag. 

**Open questions**
- [ ] Is the new ViewFilter API acceptable?
- [ ] Should `viewFilterFor` still accept a boolean lambda instead of all 3 values, since the third value is likely not needed by most filters and it would allow more readable single-expression filters.
- [ ] Is searching up the tree on every node an acceptable cost?
- [ ] This approach requires the filter to be aware of _all_ types that are visitable - a composable nested in an Android view nested inside a composable would also need to walk up the inner `View` hierarchy so that it included the complete subtrees. Would a better approach to make the filter stateful, so it could set an internal flag while visiting children that match? This would also be more efficient, since it wouldn't need to walk up the parent chain for every node. However it requires a slightly bigger change to the filter API.

One way to solve the last point would be to pass an arbitrary value down the filter chain:
```kotlin
fun interface ViewFilter<S> {
  fun matches(view: Any, filterState: S): Pair<FilterResult, S>
}

fun startFromTestTag(testTag: String): ViewFilter =
  viewFilterFor<ComposeLayoutInfo> { layoutInfo, alreadyMatched ->
    if (alreadyMatched || (testTag in testTags)) {
      Pair(INCLUDE, true)
    } else {
      Pair(INCLUDE_ONLY_CHILDREN, false)
    }
  }
```
We could also then provide two versions of `viewFilterFor`, a simple one (no filter state, boolean return) and a complete one (same signature as the above method). One issue with this is that the returned filter state value would be ignored if the result is `EXCLUDE`, so a sealed class might be better for the result than an enum at that point:
```kotlin
sealed class FilterResult<out S> {
  class Include<S>(val filterState: S) : FilterResult<S>()
  class IncludeOnlyChildren<S>(val filterState: S) : FilterResult<S>()
  object Exclude : FilterResult<Nothing>()
}
```

Fixes #58.

### DIFF
--- a/radiography/src/main/java/radiography/RenderTreeString.kt
+++ b/radiography/src/main/java/radiography/RenderTreeString.kt
@@ -1,20 +1,32 @@
 package radiography
 
+import radiography.TreeRenderingVisitor.ChildToVisit
 import radiography.TreeRenderingVisitor.RenderingScope
+import java.util.ArrayDeque
 import java.util.BitSet
 
 /**
  * Renders [rootNode] as a [String] by passing it to [visitor]'s [TreeRenderingVisitor.visitNode]
  * method.
+ *
+ * @param skip If true, this node's description will be ignored and its children will be treated
+ * as the roots.
  */
 internal fun <N> StringBuilder.renderTreeString(
   rootNode: N,
-  visitor: TreeRenderingVisitor<N>
+  visitor: TreeRenderingVisitor<N>,
+  skip: Boolean = false
 ) {
-  renderRecursively(rootNode, visitor, depth = 0, lastChildMask = BitSet())
+  renderRecursively(rootNode, visitor, skip = skip, depth = 0, lastChildMask = BitSet())
 }
 
 internal abstract class TreeRenderingVisitor<in N> {
+
+  internal data class ChildToVisit<N>(
+    val node: N,
+    val visitor: TreeRenderingVisitor<N>,
+    val skip: Boolean
+  )
 
   /**
    * Renders nodes of type [N] by rendering them as strings to [RenderingScope.description] and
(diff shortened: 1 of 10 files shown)


radiography/src/main/java/radiography/RenderTreeString.kt
```diff
@@ -57,45 +69,50 @@ private fun <N> StringBuilder.renderRecursively(
...
 
-    // Unset the bit we set above before returning.
-    if (isLastChild) {
-      lastChildMask.clear(depth)
+      // Never null on the main thread, but if called from another thread all bets are off.
+      childNode?.let {
+        renderRecursively(childNode, childNodeVisitor, depth + 1, skipChild, lastChildMask)
+      }
+    } else {
+      // Visit the child directly, discarding its description but adding its children to our queue
```
zach-klippenstein, line 107: "without generating description"

radiography/src/main/java/radiography/RenderTreeString.kt
```diff
@@ -57,45 +69,50 @@ private fun <N> StringBuilder.renderRecursively(
...
+      // to be processed by this loop.
+      val childScope = RenderingScope(null, children)
+      with(childNodeVisitor) { childScope.visitNode(childNode) }
     }
   }
+
+  // Unset the bit we might have set above before returning.
```
zach-klippenstein, line 114: "mask we set above" (remove might)

radiography/src/main/java/radiography/ViewTreeRenderingVisitor.kt
```diff
@@ -51,4 +44,21 @@ internal class ViewTreeRenderingVisitor(
...
   }
+
+  private fun RenderingScope.visitChildrenOf(view: ViewGroup) {
+    // Capture this value, since it might change while we're iterating.
+    val childCount = view.childCount
+    for (index in 0 until childCount) {
+      // Child may be null, if children were removed by another thread after we captured the child
+      // count. getChildAt returns null for invalid indices, it doesn't throw.
+      val child = view.getChildAt(index) ?: continue
+      when (viewFilter.matches(child)) {
```
zach-klippenstein, line 55: This filter+add logic is duplicated, factor out an `addFilteredChild` function?

zach-klippenstein:
Rethinking the whole codebase, might be replaced by [this branch](https://github.com/square/radiography/compare/zachklipp/rewrite).

zach-klippenstein:
This is being done a different and much better way in #70.

PR #68, opened 2020-08-29
Use a sealed class to distinguish between Android and Compose in ViewStateRenderer and ViewFilter.

zach-klippenstein:
Android-related logic is written in terms of `View`, Compose-related logic is written
in terms of `List<Modifier>`. This enables writing actual filters/renderers for Compose,
while exposing as little as possible about how Compose structures things internally.

### DIFF
--- a/radiography/src/main/java/radiography/ScannableView.kt
+++ b/radiography/src/main/java/radiography/ScannableView.kt
@@ -1,0 +1,26 @@
+package radiography
+
+import android.view.View
+import androidx.compose.ui.Modifier
+import radiography.ScannableView.AndroidView
+import radiography.ScannableView.ComposeView
+import radiography.compose.ExperimentalRadiographyComposeApi
+
+/**
+ * Represents a logic view that can be rendered as a node in the view tree.
+ *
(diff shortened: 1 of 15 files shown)


radiography/src/main/java/radiography/ScannableView.kt
```diff
@@ -0,0 +1,37 @@
...
+  @ExperimentalRadiographyComposeApi
+  public class ComposeView(
+    val width: Int,
+    val height: Int,
+    val modifiers: List<Modifier>
+  ) : ScannableView() {
+
+    internal constructor(
+      size: IntSize,
+      modifiers: List<Modifier>
```
pyricau, line 34: Does this compile without a dependency on compose? ie can you load a sealed class without loading one of the subclasses?

zach-klippenstein: Yep. It's just a generic, doesn't exist at runtime. Also for the record, we have UI tests that validate that this library works when used with an older Compose version, and with no Compose dependencies at all, so if CI is green, it's usable without Compose.

zach-klippenstein:
Updated README.

PR #69, opened 2020-08-29
Use ScannableView to walk the tree.

zach-klippenstein:
This hugely simplifies the `renderTreeString` code, removing the `TreeRenderingVisitor`
interface and replacing it with a simple function that renders a node and returns its
children. The polymorphism of tree nodes is now handled directly by `ScannableView`,
which also means we can test the tree logic separately (this PR adds a bunch of tests).

This change unblocks the introduction of `ScanScope` in a future PR (#70).

This PR is based on #68.


zach-klippenstein:
This PR is now complete and ready for review.

PR #70, opened 2020-08-31
Introduce ScanScope, a more flexible way to define what to scan.

### LINKED ISSUE #58: Drop SlotTable.scan, add composable selection support

zach-klippenstein:
Based on #69.
Closes #58.

`FocusedWindowViewFilter` is now `FocusedWindowScope`. `SlotTable.scan()` is gone.


radiography/src/main/java/radiography/Radiography.kt
```diff
@@ -40,69 +38,70 @@ public object Radiography {
   @JvmStatic
   @JvmOverloads
   public fun scan(
-    rootView: View? = null,
+    scanScope: ScanScope = AllWindowsScope,
```
pyricau, line 41: Makes me happy!

radiography/src/main/java/radiography/ScanScopes.kt
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  /**
+   * Scans all Composables in all windows that have the `Modifier.testTag` modifier with the given
+   * [testTag].
+   */
+  @ExperimentalRadiographyComposeApi
+  @JvmStatic
+  @JvmOverloads
+  fun composeTestTagScope(
```
pyricau, line 44: Might want to explain what a TestTag is and even link to the doc. https://developer.android.com/reference/kotlin/androidx/compose/ui/platform/package-summary#(androidx.compose.ui.Modifier).testTag(kotlin.String)

> Applies a tag to allow modified element to be found in tests.

It's not clear how this relates to Radiography.

pyricau: filed https://github.com/square/radiography/issues/72

zach-klippenstein: Good catch, I linked to the docs in the `ViewFilter` for test tags, but forgot to do so here.

PR #71, opened 2020-08-31
Cleanup some renderers.

### LINKED ISSUE #62: Merge compose text renderer into main one
How the contents of text views are rendered (show/hide, max length) are currently exposed in a number of ways: via multiple Defaults properties, and via parameters on text-specific renderers (both for android views and compose). These are awkward to compose: if you want to customize the text defaults, you can't use the defaults properties at all and must manually pass both the Android and the compose text renderers with the same values for their parameters.

There might be other future cases where similar renderers for related Android and compose concepts both need similar configuration, even though they're implemented as completely separate renderers.

Should we create something like a `class RenderingConfig` that would contain only the shared configuration values for stuff like text rendering? 

The text renderer factory functions would take a config object instead, and the Defaults* properties could become functions that accept an optional config object. 

Alternatively, the `ViewStateRenderer.render` method could take a config object, so the config could be changed independently from the initialization of the set of renderers and passed in through a parameter directly on the top-level scan function. 

E.g
```kotlin
// Not a data class for binary compatibility reasons
class RenderingConfig internal constructor(
  val includeText: Boolean,
  val maxTextLength: Int
)

// Can be deprecated and replaced to add values in binary compatible way. 
@JvmOverloads
fun RenderingConfig(
  includeText: Boolean = false,
  maxTextLength: Int = Int.MAX_VALUE
): RenderingConfig

fun Radiography.scan(…, config: RenderingConfig = RenderingConfig()) 

interface ViewStateRenderer {
  fun AttributeAppendable.render(
    value: Any?,
    config: RenderingConfig
  )
}
```

zach-klippenstein:
Based on #70.

- Makes the standard text value renderer also render compose text, so there's only
one place to configure text rendering.
- Merges `LayoutIdRenderer` into `ComposeViewRenderer`, since `layoutId` can be on any
Composable and there's no reason to have it separate.
- Renames the parameters to `textViewRenderer` to be more concise.

Fixes #62.


radiography/src/main/java/radiography/ViewStateRenderers.kt
```diff
@@ -55,49 +56,60 @@ public object ViewStateRenderers {
...
       CheckableRenderer,
-  ) + ComposeLayoutRenderers.DefaultsNoPii
+  )
 
   @JvmField
   public val DefaultsIncludingPii: List<ViewStateRenderer> = listOf(
       ViewRenderer,
-      textViewRenderer(includeTextViewText = true),
+      ComposeViewRenderer,
+      textViewRenderer(showTextValue = true),
```
pyricau, line 68: I think maybe the verb shouldn't be `show`, because this controls what we concatenate to a string not what we show on a string.

How about `showTextValue` ?

pyricau: and we could probably shorten to `renderText` ? idk.

pyricau: https://github.com/square/radiography/issues/73

pyricau:
Looks like my rebase broke something?

> Caused by: org.gradle.api.GradleException: There were failing tests. See the report at: file:///Users/runner/work/radiography/radiography/radiography/build/reports/androidTests/connected/index.html

radiography/src/main/java/radiography/ViewStateRenderers.kt
zach-klippenstein: I'm not sure I understand your distinction between "concatenate" and "show" in this context, but my LGAF is extremely low for this parameter. `includeTextViewText` just seemed unnecessarily long and verbose, and not strictly accurate anymore with the compose support. I'm fine with `renderText`.

zach-klippenstein:
Taking a look at the build failure.

zach-klippenstein:
Test failure is a flake, my guess is that Radiography is seeing UI from a previous test run?

```
expected to be empty
but was:
com.squareup.radiography.test/radiography.test.utilities.TestActivity:
window-focus:false
 DecorView { 1024×88px }
 `-FrameLayout { 936×0px }
   +-FrameLayout { id:content, 936×0px }
   | `-AlertDialogLayout { id:parentPanel, 936×0px }
   |   +-LinearLayout { id:topPanel, GONE, 0×0px }
   |   | +-LinearLayout { id:title_template, GONE, 0×0px }
   |   | | +-ImageView { id:icon, GONE, 0×0px }
   |   | | `-DialogTitle { id:alertTitle, 0×0px, text-length:0 }
   |   | `-Space { id:titleDividerNoCustom, GONE, 0×0px }
   |   +-FrameLayout { id:contentPanel, GONE, 0×0px }
   |   | `-ScrollView { id:scrollView, 0×0px }
   |   |   `-LinearLayout { 0×0px }
   |   |     +-Space { id:textSpacerNoTitle, 0×0px }
   |   |     +-TextView { id:message, GONE, 0×0px, text-length:0 }
   |   |     `-Space { id:textSpacerNoButtons, 0×0px }
   |   +-FrameLayout { id:customPanel, GONE, 0×0px }
   |   | `-FrameLayout { id:custom, 0×0px }
   |   `-ScrollView { id:buttonPanel, GONE, 0×0px }
   |     `-ButtonBarLayout { 0×0px }
   |       +-Button { id:button3, GONE, 0×0px, text-length:0 }
   |       +-Space { id:spacer, INVISIBLE, 0×0px }
   |       +-Button { id:button2, GONE, 0×0px, text-length:0 }
   |       `-Button { id:button1, GONE, 0×0px, text-length:0 }
   `-ViewStub { id:action_mode_bar_stub, GONE, 0×0px }

at radiography.test.RadiographyUiTest.when_noActivity_then_emptyHierarchy(RadiographyUiTest.kt:28)
```

zach-klippenstein:
Added a commit to try to fix the flake by using Android Test Orchestrator to isolate tests better.

PR #74, opened 2020-09-01
Add links and samples to documentation about test tags.

### LINKED ISSUE #72: Document TestTag
ScanScopes.composeTestTagScope() has "test tag" in its name and kdoc.

Might want to explain what a TestTag is and even link to the doc. https://developer.android.com/reference/kotlin/androidx/compose/ui/platform/package-summary#(androidx.compose.ui.Modifier).testTag(kotlin.String)

> Applies a tag to allow modified element to be found in tests.

It's not clear how this relates to Radiography.

zach-klippenstein:
Also fleshes out a lot more Compose documentation in the README, as well
as more general information about `ScanScope`.

Fixes #72.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,8 @@ val prettyHierarchy = Radiography.scan(viewStateRenderers = DefaultsNoPii +
     })
 ```
 
-You can print a subset of the view hierarchies.
+You can print a subset of the view hierarchies by specifying a `ScanScope`. By default, Radiography
+will scan all the windows owned by your app.
 
 ```kotlin
 // Extension function on View, renders starting from that view.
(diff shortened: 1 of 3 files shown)


zach-klippenstein:
@pyricau Please merge if you approve.

README.md
```diff
@@ -77,38 +78,73 @@ This sample app lives in this repo in the `sample` directory.
 
 ## Jetpack Compose support
 
-Radiography will automatically render Composables found in your view tree if the Compose Tooling
-library is on the classpath. If you're using Compose, this is likely already the case (the
+[Jetpack Compose](https://developer.android.com/jetpack/compose) is Google's new UI toolkit. It is
+currently in pre-release status, and changes frequently. However, if you're already experimenting
+with Compose in your apps, any UI that is defined in Compose will not show up as Android views.
```
pyricau, line 83: >  any UI that is defined in Compose will not show up as Android views.

this is a bit confusing, show up where exactly?

How about:

> any UI that is defned in Compose does not include any Android view.

README.md
```diff
@@ -77,38 +78,73 @@ This sample app lives in this repo in the `sample` directory.
...
-or you don't depend on the Tooling library, then Radiography will include a message in the result
-asking you to upgrade Radiography or add the Tooling library, but this is optional.
+or you don't depend on the Tooling library, then Radiography will still try to detect compositions,
+but instead of rendering the actual hierarchy, it will just include a message asking you to upgrade
+Radiography or add the Tooling library.
 
 ### Compose usage
 
-Compose support is experimental right now, since Compose is still changing a lot on a regular basis.
+_Compose support is experimental right now, since Compose is still changing a lot on a regular
```
pyricau, line 97: > changing a lot on a regular basis.

=>

> changing frequently.

?

README.md
```diff
@@ -77,38 +78,73 @@ This sample app lives in this repo in the `sample` directory.
...
+The `DefaultsNoPii` and `DefaultsIncludingPii` renderers include default Compose renderers – you
+don't need to do anything special. Additional Compose-specific renderers can be found in the
+`ComposableRenderers` object.
 
-Composables can be filtered just like views, but instead of using view ID, you can either specify
-a test tag (as in the string passed to `Modifier.testTag()`), or match on any layout ID (as in
-`Modifier.layoutId()`).
+To create a custom renderer for Compose, implement a `ViewStateRenderer` to handle values of type
+`ComposeView`. The `ComposeView` has a list of all the `Modifier`s that have been applied to the
+composable, including its [semantics](https://developer.android.com/reference/kotlin/androidx/compose/ui/semantics/SemanticsModifier?hl=en#semanticsConfiguration:androidx.compose.ui.semantics.SemanticsConfiguration).
```
pyricau, line 116: this might want a code example

README.md
```diff
@@ -77,38 +78,73 @@ This sample app lives in this repo in the `sample` directory.
...
 
-The `DefaultsNoPii` and `DefaultsIncludingPii` renderers already configure the Compose renderers as
-well. Additional compose-specific renderers can be found in the `ComposeLayoutRenderers` object.
-Custom Compose renderers cannot be created at this time, since we're still figuring out exactly
-what the best public API for that looks like. If you need a custom Compose renderer, please file
-an issue on this repo!
+To _only_ scan the contents of the `Body` composable, use `composeTestTagScope`:
+```kotlin
+val prettyHierarchy = Radiography.scan(scanScope = composeTestTagScope("app-body"))
+```
```
pyricau, line 147: The example is really cool! One thing that's not obvious when reading this is that at first it sounds like "_only_ scan the contents of the `Body`" is the opposite of "scan everything _but_ the debug drawer" so then we expect some kind of API symmetry which there isn't.

And the main reason for that is that the first is a filter vs the second is a root provider.

Might want to add an extra sentence to the second case, e.g.:

```
To _only_ scan the contents of the `Body` composable, specify a `composeTestTagScope` `ScanScope`  to tell Radiography where to start from:
```

radiography/src/main/java/radiography/ScanScopes.kt
```diff
@@ -35,8 +35,25 @@ object ScanScopes {
   }
 
   /**
-   * Scans all Composables in all windows that have the `Modifier.testTag` modifier with the given
-   * [testTag].
+   * Scans only composables that have a [`Modifier.testTag`][androidx.compose.ui.platform.testTag]
```
pyricau, line 38: > Scans only composables

This might be confusing, as it sounds like a filter when in fact it includes the entire composable hierarchy for any matched composable. 

I think the confusion comes from the fact that "scan" here means "find roots" but that's not obvious.

README.md
zach-klippenstein: Agreed it's confusing. Technically a Compose UI _may_ [include android views](https://developer.android.com/jetpack/compose/interop#views-in-compose). This whole paragraph is awkward. How about replacing it with:

> [Jetpack Compose](https://developer.android.com/jetpack/compose) is Google's new declarative UI toolkit. It is a completely new implementation, and does not use Android views itself (although it [interoperates](https://developer.android.com/jetpack/compose/interop#views-in-compose) with them seamlessly).

README.md
```diff
@@ -77,38 +78,73 @@ This sample app lives in this repo in the `sample` directory.
...
 ## Jetpack Compose support
 
-Radiography will automatically render Composables found in your view tree if the Compose Tooling
-library is on the classpath. If you're using Compose, this is likely already the case (the
+[Jetpack Compose](https://developer.android.com/jetpack/compose) is Google's new UI toolkit. It is
+currently in pre-release status, and changes frequently. However, if you're already experimenting
+with Compose in your apps, any UI that is defined in Compose will not show up as Android views.
+
+Radiography will automatically render composables found in your view tree if the Compose Tooling
+library is on the classpath. If you are using Compose, this is likely already the case (the
```
zach-klippenstein, line 86:
```suggestion
library is on the classpath. If you are using Compose, you're probably already using this library (the
```

README.md
```diff
@@ -77,38 +78,73 @@ This sample app lives in this repo in the `sample` directory.
...
 `@Preview` annotation lives in the Tooling library). On the other hand, if you're not using Compose,
 Radiography won't bloat your app with transitive dependencies on any Compose artifacts.
 
 Compose releases frequently. If you are using Radiography with an unsupported version of Compose,
```
zach-klippenstein, line 90:
```suggestion
Compose changes frequently, and while in alpha, is being released every two weeks. Radiography's Compose support is experimental. If you are using Radiography with an unsupported version of Compose,
```

README.md
zach-klippenstein: Going to remove this sentence from this paragraph and move it to the one above.

README.md
zach-klippenstein: Should these be separate samples? I was on the fence whether using the same code for both caused the confusion you pointed out.

radiography/src/main/java/radiography/ScanScopes.kt
zach-klippenstein: How about
```suggestion
   * Limits the scope of the scan to start from composables that have a [`Modifier.testTag`][androidx.compose.ui.platform.testTag]
```

zach-klippenstein: Done.

README.md
zach-klippenstein: Done.

README.md
zach-klippenstein: Done.

README.md
zach-klippenstein: Done.

README.md
zach-klippenstein: Done.

README.md
zach-klippenstein: Split them into separate samples and talked about scoping first.

README.md
zach-klippenstein: Done.

zach-klippenstein:
Addressed all comments and polished some other wordings.

PR #75, opened 2020-09-01
Remove the skipLayoutIdsFilter.

zach-klippenstein:
I think `testTag` is better-suited for this purpose, `layoutId` is really only
meant for communicating with a particular custom parent layout. But I used the
implementation for a new example in the README of how to write a custom filter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,21 @@ Custom Compose renderers cannot be created at this time, since we're still figur
 what the best public API for that looks like. If you need a custom Compose renderer, please file
 an issue on this repo!
 
+To write a custom filter for Compose, implement a `ViewFilter` to handle values of type
+`ComposeView`. For example, a filter that excludes composables with a particular `Modifier.layoutId`
+might look something like this:
+```kotlin
+fun skipLayoutIdsFilter(skipLayoutId: (Any) -> Boolean) = ViewFilter { view ->
+  (view as? ComposeView)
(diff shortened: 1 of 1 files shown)


zach-klippenstein:
This PR will cause a merge conflict with #76. If we merge #76, we don't need to merge this one, since it also removes this function.

README.md
```diff
@@ -110,6 +110,21 @@ Custom Compose renderers cannot be created at this time, since we're still figur
 what the best public API for that looks like. If you need a custom Compose renderer, please file
 an issue on this repo!
 
+To write a custom filter for Compose, implement a `ViewFilter` to handle values of type
+`ComposeView`. For example, a filter that excludes composables with a particular `Modifier.layoutId`
+might look something like this:
+```kotlin
+fun skipLayoutIdsFilter(skipLayoutId: (Any) -> Boolean): ViewFilter = ViewFilter { view ->
```
pyricau, line 117: Is `: ViewFilter` needed?

zach-klippenstein: Not technically, I'll remove it.

zach-klippenstein:
UI test failure was emulator failing to start, re-running.

zach-klippenstein:
Shards keep timing out, but it's different shards each time, so it's clear this PR is actually passing. I will merge without waiting for a fully green run.

PR #77, opened 2020-09-01
Increase timeout for running UI tests in CI.

zach-klippenstein:
From 20 -> 30 minutes. Many runs have at least one shard that times out in 20 mins.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -26,7 +26,7 @@ jobs:
   instrumentation-tests:
     name: Instrumentation tests
     runs-on: macos-latest
-    timeout-minutes: 20
+    timeout-minutes: 30
     strategy:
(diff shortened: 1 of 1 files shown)


zach-klippenstein:
Some runs are still timing out: https://github.com/square/radiography/pull/75/checks?check_run_id=1061620240

In that build, the job waits for 8 minutes _just for the emulator to start_. Gradle then takes 3 minutes after starting the daemon just to run the first task, `:buildSrc:compileKotlin`, so at the 10-minute mark, the actual code source hasn't even started compiling yet. At 20 minutes, the first test is just starting to run.

zach-klippenstein:
Happened again, on a different shard. This time installing the system image took 3 minutes, started up fairly quickly, but Gradle took 4 minutes from daemon start to finish the first task.

PR #81, opened 2020-09-03
Cleanup readme intro

pyricau:


zach-klippenstein (approved):
Nice! Should we update the tree characters here right away? We also need to update them in the blog post.

zach-klippenstein:
Updated tree characters to the fancy ones.

PR #82, opened 2020-09-03
Don't include text-length when full text included

pyricau:


zach-klippenstein (changes requested):
This seems good, but two things:

1. Test should be added for new behavior (ellipsized text that is the same length).
2. Compose text rendering behavior should match (I'm pretty sure it currently matches the TextView behavior).

zach-klippenstein (approved):

PR #85, opened 2020-09-03
Merge separate kotlin compilation configuration blocks in gradle scripts.

zach-klippenstein:
The duplicate blocks were overwriting the compiler args, so explicit API
mode was not actually being enabled.

### DIFF
--- a/radiography/src/main/java/radiography/ScanScope.kt
+++ b/radiography/src/main/java/radiography/ScanScope.kt
@@ -9,8 +9,8 @@ package radiography
  *  - [ScanScopes.FocusedWindowScope]
  *  - [ScanScopes.singleViewScope]
  */
-fun interface ScanScope {
+public fun interface ScanScope {
 
   /** Returns the [ScannableView]s that scanning should start from. */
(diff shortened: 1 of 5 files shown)


radiography/src/main/java/radiography/ScanScope.kt
```diff
@@ -9,8 +9,8 @@ package radiography
  *  - [ScanScopes.FocusedWindowScope]
  *  - [ScanScopes.singleViewScope]
  */
-fun interface ScanScope {
+public fun interface ScanScope {
```
steve-the-edwards, line 12: ? for explicitness?

steve-the-edwards (approved):

radiography/src/main/java/radiography/ScanScope.kt
zach-klippenstein: Yes, strict API mode requires all visibilities and types to be explicit. This mode had not been enabled, because we were clobbering the option to turn it on (the thing this PR fixes).

PR #88, opened 2020-09-03
Add tests in Java to validate API.

zach-klippenstein:
These tests don't run anything, but let us see how these APIs look from Java
and ensure we don't break them.


radiography/src/main/java/radiography/ViewStateRenderers.kt
```diff
@@ -183,6 +183,7 @@ public object ViewStateRenderers {
    */
   @ExperimentalRadiographyComposeApi
   @JvmStatic
+  @JvmSynthetic
   internal fun composeTextRenderer(
```
zach-klippenstein, line 187: This was showing up in Java, oops.

steve-the-edwards (approved):

PR #92, opened 2020-09-04
Fix grammatical errors

androiddevnotes:


androiddevnotes:
> These sound less grammatical to me, can you elaborate?

For both the corrections, I used Grammarly’s browser extension which makes grammar suggestions and personally the corrected sentences sound better to me as well.

README.md
```diff
@@ -133,9 +133,9 @@ The `DefaultsNoPii` and `DefaultsIncludingPii` renderers include default Compose
 don't need to do anything special. Additional Compose-specific renderers can be found in the
 `ComposableRenderers` object.
 
-To create a custom renderer for Compose, implement a `ViewStateRenderer` to handle values of type
+To create a custom renderer for Compose, implement a `ViewStateRenderer` to handle the values of type
```
pyricau, line 136: Not sure that sounds better?

README.md
```diff
@@ -133,9 +133,9 @@ The `DefaultsNoPii` and `DefaultsIncludingPii` renderers include default Compose
...
 `ComposeView`. However, since Radiography gets most of its information about composables from their
-semantics properties, in most cases you shouldn't need to define any custom rendering logic. The
+semantics properties, in most cases you shouldn't need to define any custom rendering logic.
```
pyricau, line 138: This might make sense

README.md
androiddevnotes: Added a commit to revert the changes.

README.md
zach-klippenstein: I didn't use an article initially because this sentence is saying more than the ComposeView type has properties, but I think that's wrong because in the context of the rest of the paragraph, this sentence is talking about implementations of ViewRender, and in the context of a particular implementation, it makes more sense to talk about instance vs type. However, I think it's better to use a plural here to match above plural "values if type ComposeView", for consistency. Does that make sense?

zach-klippenstein:
UI test timed out, but this is just a docs change so I'm gonna merge anyway.

PR #94, opened 2020-09-07
Sort skip ids before using in binary search

samruston:
`skipIdsViewFilter` is using a binary search without making sure the view ids are sorted. This PR simply sorts it when creating the view filter. Added a test that previously failed because the list is unsorted.


steve-the-edwards (approved):

zach-klippenstein (approved):
Thanks for doing this!

Issue #93, opened 2020-09-06
Scanning stops at subcomposition boundaries


zach-klippenstein:
E.g. `WithConstraints` will always be a leaf.

zach-klippenstein:
CompositionReference is stored in the child Composer, and the child is registered on the parent. So I think we can get the list of all subcompositions, then we just need to note when we see a composition reference in the slot table, and look up the corresponding subcomposition to get its children.

zach-klippenstein:
`Scaffold` is being refactored to use subcomposition. This feature is about to become a lot more important.

https://android-review.googlesource.com/c/platform/frameworks/support/+/1426753

zach-klippenstein:
I [think](https://twitter.com/zachklipp/status/1310806712614486017?s=19) the best way to do this is to specifically look for SubcomposeLayouts and composition references inside them, and then treat all subcomposition nodes as immediate children of the SubcomposeLayout. This should cover almost all the actual cases of subcompositions being visual children. The only cases I can think of that it doesn't cover are either Dialogs (in which case the dialog isn't visually a child of wherever the dialog composable is called, and will be scanned as a separate window anyway), and weird fancy stuff like the proposed workflow view binding composition linking which always has classic view trees in the middle anyway.

PR #95, opened 2020-09-09
Draw nested logo effect programmatically in Compose sample.

zach-klippenstein:
This uses subcomposition and doesn't scan correctly due to #93.

image


steve-the-edwards (approved):

sample-compose/src/main/java/com/squareup/radiography/sample/compose/RadiographyLogo.kt
```diff
@@ -0,0 +1,91 @@
...
+  val outerImage = imageResource(R.drawable.logo_outer)
+  val aspectRatio = outerImage.width.toFloat() / outerImage.height.toFloat()
+
+  Stack(modifier.aspectRatio(aspectRatio)) {
+    Image(outerImage)
+    InfiniteMirror(
+        centerOffsetFraction = Offset(.5f, .59f),
+        scaleFactor = .46f
+    ) {
+      Image(imageResource(R.drawable.logo_inner))
```
zach-klippenstein, line 40: imageResource should be pulled out of this lambda so it's not being loaded at every level.

PR #98, opened 2020-09-17
Update the compose sample to use the latest compose version.

zach-klippenstein:
Note that this does not actually change the version of compose that the library is compiled against. This ensures that the existing release still works with the newer compose version.

### DIFF
--- a/sample-compose/build.gradle.kts
+++ b/sample-compose/build.gradle.kts
@@ -6,7 +6,7 @@ plugins {
 }
 
 /** Use a separate property for the sample so we can test with different versions easily. */
-val sampleComposeVersion = Versions.Compose
+val sampleComposeVersion = "1.0.0-alpha03"
 
 android {
   compileSdkVersion(30)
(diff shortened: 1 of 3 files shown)


pyricau (approved):

sample-compose/build.gradle.kts
```diff
@@ -6,7 +6,7 @@ plugins {
 }
 
 /** Use a separate property for the sample so we can test with different versions easily. */
-val sampleComposeVersion = Versions.Compose
+val sampleComposeVersion = "1.0.0-alpha03"
```
pyricau, line 9: could also extract this as "Versions.MinSupportedCompose" ?

zach-klippenstein: Sure, although it would be `MaxSupportedCompose`. Maybe we should extract min/max versions and run both in CI?

zach-klippenstein: #100

patrickyin (approved):

PR #113, opened 2020-11-25
Upgrade Kotlin to 1.4.20 and remove the older API level

vRallev:


vRallev:
Kinda annoying setup that Square members don't have push permissions and must create a fork.

.github/workflows/android.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         kotlin-version:
-          - 1.4.0
```
vRallev, line 15: Probably that was forgotten to bump.

build.gradle.kts
```diff
@@ -78,8 +77,6 @@ subprojects {
       }
 
       jvmTarget = "1.6"
-      // Required to avoid warnings about using older stdlib version.
-      apiVersion = "1.3"
```
vRallev, line 82: This was probably only needed for Register, so I removed it.

zach-klippenstein:
Unfortunately I don't think this will work as-is because Compose doesn't support 1.4.20 yet. It should in the next release, and I was planning to update this repo then.

vRallev:
Oh, I think Kotlin `1.4.20` requires a new Compose version. I created a separate PR for the first commit: #114. I leave the Kotlin 1.4.20 upgrade up to you.

PR #114, opened 2020-11-25
Remove the API 1.3 workaround. 

vRallev:
It's not longer needed and the latest standard library can be used.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -19,7 +19,6 @@ import org.jlleitschuh.gradle.ktlint.KtlintExtension
 import org.jlleitschuh.gradle.ktlint.reporter.ReporterType
 
 println("Building with Kotlin compiler version ${Versions.KotlinCompiler}")
-println("Building with Kotlin stdlib version ${Versions.KotlinStdlib}")
 
 buildscript {
   repositories {
@@ -78,8 +77,6 @@ subprojects {
       }
 
       jvmTarget = "1.6"
-      // Required to avoid warnings about using older stdlib version.
-      apiVersion = "1.3"
     }
   }
 
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -5,9 +5,6 @@ object Versions {
    */
   val KotlinCompiler = System.getProperty("square.kotlinVersion") ?: "1.4.10"
 
-  /** Use a lower version of the stdlib so the library can be consumed by lower kotlin versions. */
-  val KotlinStdlib = System.getProperty("square.kotlinStdlibVersion") ?: "1.3.72"
-
   const val Compose = "1.0.0-alpha05"
 }
 
--- a/compose-tests/build.gradle.kts
+++ b/compose-tests/build.gradle.kts
@@ -35,9 +35,6 @@ android {
 tasks.withType<KotlinCompile> {
   kotlinOptions {
     jvmTarget = "1.8"
-    // Override lower API version from root build.gradle.kts since this module is using the new
-    // stdlib.
-    apiVersion = "1.4"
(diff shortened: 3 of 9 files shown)


build.gradle.kts
```diff
@@ -78,8 +77,6 @@ subprojects {
       }
 
       jvmTarget = "1.6"
-      // Required to avoid warnings about using older stdlib version.
-      apiVersion = "1.3"
```
vRallev, line 82: I assume that was only needed for Register. Not anymore.

zach-klippenstein:
This was required for register, but changing this is a change that potentially breaks backwards-compatibility for other consumers, so it requires us to bump the major version. @pyricau should we do that and drop Kotlin 1.3.x support going forward, now that we don't need it internally?

vRallev:
Note that OkHttp for example doesn't bump the major version either: https://github.com/square/okhttp/pull/6341

zach-klippenstein:
Semver states changes that break backwards compatibility must bump major version number. I don't know if OkHttp is using semver, but this project is.

rjrjr:
It's just a number. Bump it!

pyricau:
* Does this change have the potential to create issues for consumers that otherwise use Kotlin 1.3?
  * If yes, does this change solve an issue for consumers that use Kotlin 1.4?
    * If that's a strong enough yes, it's worth merging this but since we answer yes to "upgrading causes issues" then it's obviously a breaking change for consumers.
    * If no, does this change significantly improve the life of maintainers of Radiography? (I think no?) => probably shouldn't make this change.
  * if no, we can merge this without a major bump. I doubt the answer is no, otherwise why would we even have had this in the first place.

pyricau:
Following the line of questioning above I'm tempted to say we shouldn't undo this, but maybe I don't understand what's changed with the situation here. This library isn't internal so an internal Square change shouldn't drive a change here.

vRallev:
You had this mainly because our internal codebase is huge and we couldn't upgrade Kotlin for a few months. That has been resolved. If we wouldn't have had this internal issue, then you wouldn't have changed the api version at all. No other library is doing that: OkHttp, Workflow, Anvil, AndroidX, etc. 

I strongly believe the answer to all of that is No and it can safely be removed.

vRallev:
> This library isn't internal so an internal Square change shouldn't drive a change here.

Without the internal conflicts the API version wouldn't have been changed in the first place.

pyricau:
Right, I'm saying, if we had this issue then other consumers likely have the exact same issue. If there's no significant benefit to upgrading AND upgrading forces some of the consumers who update to do significant migration work, then they likely just won't update Radiography. i.e. we'd be increasing friction to adoption of new versions of Radiography. And so my question is: is it worth it?

vRallev:
I believe it's not such a big deal as you think. It's not problem for OkHttp with way more consumers and it won't be a problem for Radiography. The whole industry should move forward and not stay behind and it's certainly a nice forcing function when libraries upgrade. To answer your question: No, in my opinion it's not worth to keep backwards compatibility for Kotlin 1.3 in this library. Especially since there are references to Compose, which imply the new IR backend and a Kotlin version higher than 1.4.20 at some point.

zach-klippenstein:
- I don't think upgrading to 1.4.20 significantly improves the maintenance burden at this time. 
- It may increase friction and make adoption harder (although I agree with Ralf - a couple months after the 1.4 release, if we were able to upgrade internally i suspect everyone else already has).
- If we were to bump the major version and upgrade, consumers could still use an older version of the library if it's a real issue for them. 
- I think my initial caution was unwarranted, let's bump the major and upgrade.

steve-the-edwards (approved):
Looking at the discussion seems as though we concluded OK to. move forward but never hit the button.

PR #117, opened 2021-01-07
Upgrade Compose to 1.0.0-alpha09.

zach-klippenstein:


buildSrc/src/main/java/Dependencies.kt
```diff
@@ -3,9 +3,9 @@ object Versions {
    * To change this in the IDE, use `systemProp.square.kotlinVersion=x.y.z` in your
    * `~/.gradle/gradle.properties` file.
    */
-  val KotlinCompiler = System.getProperty("square.kotlinVersion") ?: "1.4.10"
+  val KotlinCompiler = System.getProperty("square.kotlinVersion") ?: "1.4.21"
```
zach-klippenstein, line 6: Latest compose version requires this kotlin version.

compose-tests/src/androidTest/java/radiography/test/compose/ComposeUiTest.kt
```diff
@@ -46,15 +45,15 @@ import radiography.ViewStateRenderers.textViewRenderer
...
-@OptIn(ExperimentalRadiographyComposeApi::class, ExperimentalSubcomposeLayoutApi::class)
+@OptIn(ExperimentalRadiographyComposeApi::class)
 class ComposeUiTest {
 
   @get:Rule
   val composeRule = createComposeRule()
 
   @Test fun when_includingPii_then_hierarchyContainsText() {
     composeRule.setContentWithExplicitRoot {
-      Text("FooBar")
```
zach-klippenstein, line 57: `Text` is now material-specific.

compose-tests/src/androidTest/java/radiography/test/compose/ComposeUiTest.kt
```diff
@@ -498,10 +497,12 @@ class ComposeUiTest {
...
-        LazyColumnFor(items = listOf(1, 2, 3), modifier = Modifier.testTag("list")) {
-          Box(Modifier.testTag("child:$it"))
-          if (it % 2 == 0) {
-            Box(Modifier.testTag("child:$it (even)"))
+        LazyColumn(Modifier.testTag("list")) {
+          items(listOf(1, 2, 3)) {
+            Box(Modifier.testTag("child:$it"))
+            if (it % 2 == 0) {
+              Box(Modifier.testTag("child:$it (even)"))
+            }
```
zach-klippenstein, line 505: Our code style changed, probably want to do an indentation pass over the whole codebase.

compose-tests/src/androidTest/java/radiography/test/compose/ComposeUiTest.kt
```diff
@@ -573,9 +574,9 @@ class ComposeUiTest {
     modifier: Modifier,
     children: @Composable () -> Unit
   ) {
-    SubcomposeLayout<Unit>(modifier) { constraints ->
+    SubcomposeLayout(modifier) { constraints ->
```
zach-klippenstein, line 577: This isn't being suddenly inferred, the type parameter was actually removed.

radiography/src/main/java/radiography/internal/ComposeLayoutInfo.kt
```diff
@@ -50,18 +51,19 @@ private fun Group.computeLayoutInfos(
...
-        )
-      }
+    .flatMap { it.tryGetComposers().asSequence() }
+    .map { subcomposer ->
+      ComposeLayoutInfo(
+        isSubcomposition = true,
+        name = name,
+        bounds = box,
+        modifiers = emptyList(),
+        children = subcomposer.compositionData.asTree().layoutInfos,
```
zach-klippenstein, line 63: `slotTable` is now private, replaced by these more limited-surface-area APIs. They're still not intended for public consumption, hence the need for the `@OptIn` above.

compose-tests/build.gradle.kts
```diff
@@ -30,6 +30,11 @@ android {
     kotlinCompilerVersion = Versions.KotlinCompiler
     kotlinCompilerExtensionVersion = Versions.Compose
   }
+
+  packagingOptions {
```
rjrjr, line 34: Could you add a comment about why this is needed?

radiography/src/main/java/radiography/internal/ComposeLayoutInfo.kt
rjrjr: Worth saying so in a code comment, that `compositionData` specifically is "cheating."

rjrjr (approved):

compose-tests/build.gradle.kts
zach-klippenstein: It's the same reason it's always needed for most kotlin libraries, I can link to more info but it's pretty standard.

zach-klippenstein:
The build is failing because the GitHub workflow overrides the kotlin version to an old one: https://github.com/square/radiography/blob/2e0dff64cf4b9a4ea3bd559f9f800ba18628902b/.github/workflows/android.yml#L15

@pyricau do you remember why that's there? I think we should remove it.

compose-tests/src/androidTest/java/radiography/test/compose/ComposeUiTest.kt
zach-klippenstein: #118

PR #120, opened 2021-01-22
Start running UI tests on API 30 emulators in CI.

zach-klippenstein:


zach-klippenstein:
Ah, the `default` target system image for API 30 isn't available yet (see https://github.com/ReactiveCircus/android-emulator-runner/issues/106#issuecomment-729974225). We need to use `google_apis` instead.

radiography/build.gradle.kts
```diff
@@ -22,9 +22,7 @@ plugins {
 }
 
 android {
-  // Using 28 for now to get the Robolectric tests working.
   compileSdkVersion(30)
-  buildToolsVersion = "29.0.2"
```
zach-klippenstein, line 27: This has been optional since AGP 3, and so instead of upgrading this to the latest version I just removed it.

radiography/src/androidTest/java/radiography/test/RadiographyUiTest.kt
```diff
@@ -19,36 +18,33 @@ import java.util.concurrent.TimeUnit.SECONDS
 
 class RadiographyUiTest {
 
-  @get:Rule
-  val activityRule = ActivityTestRule(TestActivity::class.java, false, false)
```
zach-klippenstein, line 23: This is now deprecated, replaced by `ActivityScenario` and `ActivityScenarioRule` (we can't use the rule because each test passes a different intent).

sample/build.gradle.kts
```diff
@@ -43,6 +43,7 @@ dependencies {
   androidTestImplementation(Dependencies.InstrumentationTests.Core)
   androidTestImplementation(Dependencies.InstrumentationTests.Espresso)
   androidTestImplementation(Dependencies.InstrumentationTests.Rules)
+  androidTestImplementation(Dependencies.InstrumentationTests.JUnit)
```
zach-klippenstein, line 46: Needed for `ActivityScenarioRule`.

pyricau (approved):

zach-klippenstein:
Tests seem to be flaking on 23 and 26:
```
radiography.test.RadiographyUiTest > when_includingPii_then_hierarchyContainsText[test(AVD) - 8.0.0] FAILED 
	expected to contain: Yo
	but was            :
```

pyricau (approved):

PR #124, opened 2021-02-06
Fix UI Tests reports

pyricau:
* Enable result uploading on failure (default is only success)
* Rename uploaded artifact so that we get one file per API level instead of overwriting the same file N times.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -50,7 +50,8 @@ jobs:
           script: ./gradlew connectedCheck --no-build-cache --no-daemon --stacktrace
 
       - name: Upload results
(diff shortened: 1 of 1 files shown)


.github/workflows/android.yml
```diff
@@ -50,7 +50,8 @@ jobs:
           script: ./gradlew connectedCheck --no-build-cache --no-daemon --stacktrace
 
       - name: Upload results
+        if: ${{ always() }}
```
zach-klippenstein, line 53: Oh wow, good catch. Gonna go fix this everywhere…

zach-klippenstein (approved):

PR #127, opened 2021-02-28
Upgrade AGP and gradle to latest versions.

zach-klippenstein:
Had to raise the Compose version for unsuppported tests because the IR compiler
was somehow not happy with the newer AGP.


zach-klippenstein:
Feel free to merge on approval. I'll merge sometime on Sunday if nobody's looked at it, since this is just tooling upgrades.

PR #128, opened 2021-02-28
 Upgrade Compose to beta01.

zach-klippenstein:
There were some actual changes required to this – most notably:
- The way that Android views are embedded in compositions changed quite a bit so that required some rewriting.
- `CompositionReference` was renamed to `CompositionContext`, so I renamed our related definitions to match.
- `TextField`s now report their text via the `EditableText` property, the `Text` property is reserved for non-editable text. There were also a lot more semantic properties added, so we now report all of those.

This PR updates the unsupported-version test to Compose alpha12, which means we will probably report something at vaguely useful from that version, but not necessarily any earlier. Compose being in beta means that it is now API-stable, which means that we should be able to continue to support much earlier versions of compose much longer – we just have to internally be aware if different Compose-internal implementation details change. So far we usually end up having to bump the minimum version almost every release because it was just too much work to support internal differences _and_ API differences.


pyricau (approved):

zach-klippenstein:
This is the second time in a row these tests failed - previously on a PR that didn't change any code. Either they're flaky, or some infra broke recently:
```

radiography.test.RadiographyUiTest > when_showDialog_then_hierarchyHasTwoWindows[test(AVD) - 11] FAILED 
529
	expected to be true
530
	at radiography.test.RadiographyUiTest.waitForFocus(RadiographyUiTest.kt:142)
531

532
radiography.test.RadiographyUiTest > when_onlyFocusedWindow_then_hierarchyHasOnlyDialog[test(AVD) - 11] FAILED 
533
	expected to be true
534
	at radiography.test.RadiographyUiTest.waitForFocus(RadiographyUiTest.kt:142)
535

```

PR #135, opened 2021-03-02
Refresh live compose scan by observing RecomposerInfo instead of only the top-most recompose scope.

zach-klippenstein:


pyricau:
Is there a "I have no idea what YOU're doing" dog meme?

sample-compose/src/main/java/com/squareup/radiography/sample/compose/ComposeSampleApp.kt
```diff
@@ -131,12 +145,8 @@ internal const val LIVE_HIERARCHY_TEST_TAG = "live-hierarchy"
...
       }
 
-      SideEffect {
-        liveHierarchy.value = Radiography.scan(
-          viewStateRenderers = DefaultsIncludingPii,
-          // Don't trigger infinite recursion.
-          viewFilter = skipComposeTestTagsFilter(LIVE_HIERARCHY_TEST_TAG)
-        )
+      LaunchedEffect(Unit) {
+        runLiveScanning { liveHierarchy.value = it }
```
pyricau, line 149: cool so this effectively says "rescan every time we've recomposed". btw does that not retrigger a composition / any risk of infinite loop?

pyricau (approved):

zach-klippenstein:
> Is there a "I have no idea what YOU're doing" dog meme?

image

sample-compose/src/main/java/com/squareup/radiography/sample/compose/ComposeSampleApp.kt
zach-klippenstein: Correct. It does not trigger infinite composition loop, although I'm not super clear on why not. I can look into it and document, because it does seem like that should happen.

zach-klippenstein: Ok I get why there's no infinite loop now. Because we exclude the scan result composable from the scan, a scan in which only the scan output changes will result in the scan string being set to the same value, which won't invalidate any recompose scopes.

steve-the-edwards (approved):

PR #136, opened 2021-03-02
Use Curtains instead of WindowScanner

pyricau:
This allows us to centralize the hacky code into a neat hack library so that future hacky tweaks happen only once.


zach-klippenstein (approved):
🎉 🚀

pyricau:
Oh I just realized I should mention Curtain enforces that you're retrieving the root views from the main thread.

zach-klippenstein (approved):

PR #140, opened 2021-08-12
Upgrade Gradle and AGP versions to 7.1.1 and 7.0.0, respectively.

zach-klippenstein:


zach-klippenstein:
The SDK 29 shard keeps failing because the emulator can't be connected to, I think we can just ignore that.

PR #149, opened 2022-02-11
Updated README.md

lpizzinidev:
Updated tooling library to the latest version in README file


steve-the-edwards:
This is not a dependency that needs upgrading, its a sample in a README that is actually better as a generalized reference.

steve-the-edwards (changes requested):
Should be left as general as it is documentation reference and a specific version will get stale very quickly.

PR #154, opened 2023-04-28
Update compose library versions

efrohnhoefer:


radiography/api/radiography.api
```diff
@@ -93,3 +104,79 @@ public final class radiography/ViewsKt {
 	public static synthetic fun scan$default (Landroid/view/View;Ljava/util/List;Lradiography/ViewFilter;ILjava/lang/Object;)Ljava/lang/String;
 }
 
+public abstract class radiography/internal/ComposeLayoutInfo {
```
pyricau, line 107: How comes the API file has so much new content? Doesn't look like there's any code change?

PR #160, opened 2024-04-01
Expose call group chain information

elihart:
Our app relies heavily on lots of decomposition within Compose UI in order to organize UI and promote code reuse. However, this means we have lots of Compose functions which are basically wrapper functions and there can be quite a lot of functional nesting like that.

I was quite confused for a while about why these functions didn't show up in the Radiography output, or would show up sporadically. If we have a ui module that exports a compose "component" we want that component name to identify the usage of the component in the radiography output. However, currently this doesn't reliably happen and instead we mostly see more primitive elements show up in the output scan.

The code comment on `computeLayoutInfos` thankfully explains why this happens - only the nodes that emit layouts are represented in the final output, and they inherit the name of the top most call group wrapping them. I can see why this was done, to prevent very verbose chains, but unfortunately it doesn't always result in the best output.

This PR tries to expose more granular information about these call chains so that users can optionally get more insight into what function calls were made. I also included the source location of these function calls so that our tooling can make use of that to make more educated guesses about which functions to surface (eg, if we know its a file in our project, include the call name). To acces this, there is now a property `val callChain: List<CallGroupInfo>` on ComposeView which can be used to get information about all of the call groups wrapping the layout node.

Usage can be seen inside the test: https://github.com/square/radiography/pull/160/files#diff-0a37f7ae94183d0cefb9e8ba89d53a9409011f55230e69ad4300cacee413734eR161

The original behavior in terms of which display names are used is unchanged, and in general there is no impact to existing users. This can be seen in the screenshot of the compose app before and after, and all existing tests pass without changes needed. I did add a test to verify the call chain information is captured correctly.

I hope that this additional capability is welcome, but I am open to suggestions about how to modify the approach to better fit the direction of the library.

Before:
radiography before

After: (note there are no differences)
radiograpy after

### DIFF
--- a/compose-tests/src/androidTest/java/radiography/test/compose/ComposeViewTest.kt
+++ b/compose-tests/src/androidTest/java/radiography/test/compose/ComposeViewTest.kt
@@ -135,7 +135,7 @@ class ComposeViewTest {
       .isEqualTo(0)
 
     assertThat(callGroup.displayName)
-      .isEqualTo("Call3")
+      .isEqualTo("Call1")
 
     assertThat(callGroup.semanticsConfigurations.firstNotNullOfOrNull { it[SemanticsProperties.Text] }
       ?.map { it.toString() })
@@ -157,7 +157,7 @@ class ComposeViewTest {
       .hasSize(6)
 
     assertThat(callGroup.callChain.map { it.name })
-      .containsExactly("Call3", "Call2", "Call1", "BasicText", "Layout", "ReusableComposeNode")
+      .containsExactly("Call1", "Call2", "Call3", "BasicText", "Layout", "ReusableComposeNode")
 
     assertThat(callGroup.callChain.map { it.location?.sourceFile })
       .containsExactly(
@@ -173,23 +173,24 @@ class ComposeViewTest {
   private fun renderCallChainUI() {
     composeRule.setContentWithExplicitRoot {
 
+
       @Composable
-      fun Call1() {
+      fun Call3() {
         BasicText(text = "hello")
       }
 
       @Composable
       fun Call2() {
-        Call1()
+        Call3()
       }
 
       @Composable
-      fun Call3() {
+      fun Call1() {
         Call2()
       }
 
       Column {
-        Call3()
+        Call1()
       }
     }
   }
--- a/compose-tests/src/androidTest/java/radiography/test/compose/ComposeViewTest.kt
+++ b/compose-tests/src/androidTest/java/radiography/test/compose/ComposeViewTest.kt
@@ -8,13 +8,16 @@ import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.text.BasicText
 import androidx.compose.material.Button
+import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.layout.LayoutModifier
 import androidx.compose.ui.layout.Measurable
 import androidx.compose.ui.layout.MeasureResult
 import androidx.compose.ui.layout.MeasureScope
 import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.semantics.SemanticsProperties
 import androidx.compose.ui.test.junit4.createAndroidComposeRule
+import androidx.compose.ui.tooling.data.UiToolingDataApi
 import androidx.compose.ui.unit.Constraints
 import androidx.compose.ui.unit.Density
 import androidx.compose.ui.unit.dp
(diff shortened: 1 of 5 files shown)


rjrjr:
So if I'm following you, the logs should be different if `@OptIn(UiToolingDataApi::class)` is used? Could you add an example of that to the commit message?

elihart:
@rjrjr not quite, the current logs should not change at all, and current users won't need to opt in to UiToolingDataApi::class

What is new, is that there is now a property `val callChain: List<CallGroupInfo>` on ComposeView which can be used to get information about all of the call groups wrapping the layout node.

Usage can be seen inside the test https://github.com/square/radiography/pull/160/files#diff-0a37f7ae94183d0cefb9e8ba89d53a9409011f55230e69ad4300cacee413734eR161

If someone does want to use the sourceLocation property they will need to opt in to UiToolingDataApi::class. I considered that we may want to create a wrapper class to avoid this, but I think directly exposing this tooling class is fine because if it does change we would need to change the wrapper anyway in potentially breaking ways, and having the user opt in makes them aware of the source of the data and that it is potentially unstable.

rjrjr:
LGTM (and thanks!), but could you put that explanation in the kdoc for `UiToolingDataApi`?

compose-tests/src/androidTest/java/radiography/test/compose/ComposeViewTest.kt
```diff
@@ -109,6 +117,53 @@ class ComposeViewTest {
...
+    assertThat(callGroup.displayName).isEqualTo("Call3")
+    assertThat(callGroup.semanticsConfigurations.firstNotNullOfOrNull { it[SemanticsProperties.Text] }?.map { it.toString() })
+      .containsExactly("hello")
+
+    assertThat(callGroup.callChain).hasSize(6)
+    assertThat(callGroup.callChain.map { it.name })
+      .containsExactly("Call3", "Call2", "Call1", "BasicText", "Layout", "ReusableComposeNode")
+    assertThat(callGroup.callChain.map { it.location?.sourceFile })
+      .containsExactly("ComposeViewTest.kt", "ComposeViewTest.kt", "ComposeViewTest.kt", "ComposeViewTest.kt", "BasicText.kt", "Layout.kt")
+  }
```
pyricau, line 165: This test seems to be testing several things (at least 2) and should be broken up to test one thing per test.

But also, this isn't testing any Radiography code, right? This test is making sure that the Compose behavior doesn't change? I don't know if this should exist here. Maybe yes, but then it should probably be in a separate test class since it actually tests compose.

pyricau:
Would you mind updating the README with examples of how to get one output vs the other more detailed output? I'm a little confused.

rjrjr:
> @rjrjr `UiToolingDataApi` is part of the androidx compose tooling library. Is there somewhere else you want me to put the documentation? I can update the readme as Py requested.

Nope, nevermind, I was being an idiot. I could have sworn I saw you introducing it in this PR.

elihart:
@pyricau I've split the test and updated the readme with explanations and examples

rjrjr (approved):
This is marvelous, thank you so much.

elihart:
@rjrjr The instrumented tests for API 30 seem to be failing, but I believe it is unrelated to my changes. It looks like the last commit to main also failed like this

image

rjrjr:
I too can't reproduce this problem locally, on `main` or on Eli's branch. And I just had a green run of the test on another PR branch.

I do notice that the tests are taking a _very_ long time to run, especially including AVD startup; and that `(30)` takes the longest. @RBusarow I'm inclined to try putting Radiography on the same paid testing set up that Workflow is on. The rate of change here is so low it's hard to imagine that being an issue. WDYT?

rjrjr:
> I too can't reproduce this problem locally, on `main` or on Eli's branch. And I just had a green run of the test on another PR branch.
> 
> I do notice that the tests are taking a _very_ long time to run, especially including AVD startup; and that `(30)` takes the longest. @RBusarow I'm inclined to try putting Radiography on the same paid testing set up that Workflow is on. The rate of change here is so low it's hard to imagine that being an issue. WDYT?

Experimenting here: https://github.com/square/radiography/pull/161

rjrjr:
It's very clearly a flake predating this work. I'm going to go ahead and merge it. And then we can play the "how do we release this thing again?" game!

RBusarow:
> @RBusarow I'm inclined to try putting Radiography on the same paid testing set up that Workflow is on. The rate of change here is so low it's hard to imagine that being an issue. WDYT?

I completely agree. Good idea!

elihart:
A release would be great @rjrjr ! Thanks for following up :)

PR #163, opened 2024-04-22
Use workflow-kotlin-test-runner-ubuntu-4core to unflake

rjrjr:
Updates our GHA config to take advantage of the runner that was created to unflake Workflow's integration tests, for the same reason.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -1,18 +1,16 @@
 name: Android CI
 
-on:
-  push:
-    branches: [ main ]
-  # Build on all pull requests, regardless of target.
-  pull_request:
+on :
+  pull_request :
+  merge_group :
 
 jobs:
   build:
     strategy:
       fail-fast: false
-    runs-on: ubuntu-latest
+    runs-on : workflow-kotlin-test-runner-ubuntu-4core
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/android.yml
```diff
@@ -1,18 +1,16 @@
 name: Android CI
 
-on:
-  push:
-    branches: [ main ]
-  # Build on all pull requests, regardless of target.
-  pull_request:
+on :
+  pull_request :
+  merge_group :
```
handstandsam, line 5: Does this need to be changed as part of this?

steve-the-edwards: It depends on how securely we want to lock down the custom runner. if we specify the workflows it can run on rather than just open it to any workflow then we need a stable workflow name - which can only happen if we run it on the `main` branch rather than the new branch that is created for each PR.

steve-the-edwards:
We need to add the KVM group perms to enable HW acceleration for teh emulator in the action.

See https://github.com/square/workflow-kotlin/pull/1147/files?short_path=6b8d3aa#diff-6b8d3aa6fcc54e0e940045c4c73d85625e9f4ecf8da7a5841b17ca5210bd9cdb

Issue #159, opened 2024-02-06
Function names lost with 1.6


pmecho:
When using Compose 1.6.0, the function names are no longer available. This is reproduced in the `nestedLayouts` test which fails: 
```
expected to contain:
Box:
 Box { test-tag:"root" }
 ├─Box
 ├─Column
 │ ├─Box
 │ ╰─Box
 ╰─Row
   ├─Box
   ╰─Box
but was:
:
  { test-tag:"root" }
 ├─
 ├─
 │ ├─
 │ ╰─
 ╰─
   ├─
   ╰─

at radiography.test.compose.ComposeUiTest.nestedLayouts(ComposeUiTest.kt:350)
```

It _looks_ to be related to this change in Compose: https://android-review.googlesource.com/c/platform/frameworks/support/+/2741462. However, it doesn't seem as simple as calling `composer.collectParameterInformation()`.

pmecho:
That test passes with 1.6.0-alpha05 but fails in alpha06. The alpha06 build was the first to contain that change in the description so it does seem more likely that that is the issue: https://android.googlesource.com/platform/frameworks/support/+log/9ce75ea1baa2af10dda8f787f69c40c6654bcac1..91bb8c1f81dcb031bda00fcd8e55f1e890b56f06/compose/ui

pmecho:
It looks like you can access the composer via the `androidx.compose.runtime.currentComposer` global val when in composition. A workaround would be to call `currentComposer.collectParameterInformation()` in the `setContent` block prior to adding any composables. I'm not quite sure what the holistic fix for this is but a few options come to mind:
- Document that this call must be made and add it to all tests
- Create a custom rule that calls this automatically

I wonder if this global val could also replace the reflection done in `ComposeViews.tryGetLayoutInfos` to get the composer. Though I'm not sure the level of refactoring that would need to be done to do that as that would require a composable scope.

zach-klippenstein:
Discussed a bit offline. Radiography doesn't directly insert itself into the composition, so it's not easy to get the "current" composer in the general case. However, for this specific use case, inside a test, Radiography could probably provide a helper that would set things up correctly.

pyricau:
How could we make this work beyond tests? Can we grab the component from the compose view?

Also, @zach-klippenstein could this use case maybe be considered by the Compose team :) ? Is it really unreasonable to need access to an API that allows us to describe a compose hierarchy and export that somewhere?

justin-fiedler:
[...]

PR #169, opened 2024-07-17
Support Compose 1.6 and 1.7

elihart:
This PR makes a few updates to get all tests passing on the latest compose 1.7 beta version, addressing some changes for 1.6 and 1.7.

First, it addresses an issue due to 1.6 (https://github.com/square/radiography/issues/159) by adding calls to `currentComposer.collectParameterInformation()` within test functions. This fixes tests within Radiography to make the project compatible, but does not address the potential of providing some functionality to end users to apply this easily. It is a fairly easy function to add, so perhaps the documentation should just be updated to included a note about this (I didn't attempt to do that in this PR though)

The other main fix is to the `findTestTags` function and compose renderer - these broke because they were trying to read semantic information from the compose modifiers; in newer versions of compose the semantics information is only present in semantic nodes. I had previously updated Radiography to expose that data for Compose 1.5 (https://github.com/square/radiography/pull/157) but these usage sites hadn't been updated and had broken.

### DIFF
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -6,7 +6,9 @@ object Versions {
   val KotlinCompiler = System.getProperty("square.kotlinVersion") ?: "1.9.10"
 
   const val AndroidXTest = "1.5.0"
-  const val Compose = "1.5.3"
+//  const val Compose = "1.5.3"
+//  const val Compose = "1.6.8"
+  const val Compose = "1.7.0-beta05"
 }
 
 object Dependencies {
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -9,6 +9,7 @@ object Versions {
 //  const val Compose = "1.5.3"
 //  const val Compose = "1.6.8"
   const val Compose = "1.7.0-beta05"
+  const val ComposeCompiler = "1.5.3"
 }
 
 object Dependencies {
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -6,8 +6,6 @@ object Versions {
   val KotlinCompiler = System.getProperty("square.kotlinVersion") ?: "1.9.10"
 
   const val AndroidXTest = "1.5.0"
-//  const val Compose = "1.5.3"
-//  const val Compose = "1.6.8"
   const val Compose = "1.7.0-beta05"
   const val ComposeCompiler = "1.5.3"
 }
--- a/compose-tests/src/androidTest/java/radiography/test/compose/ComposeUiTest.kt
+++ b/compose-tests/src/androidTest/java/radiography/test/compose/ComposeUiTest.kt
@@ -14,6 +14,7 @@ import androidx.compose.material.Button
 import androidx.compose.material.Checkbox
 import androidx.compose.material.TextField
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.currentComposer
 import androidx.compose.ui.ExperimentalComposeUiApi
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.layout.SubcomposeLayout
@@ -380,6 +381,8 @@ class ComposeUiTest {
 
   @Test fun scanningHandlesDialog() {
     composeRule.setContent {
+      currentComposer.collectParameterInformation()
+
       Box(Modifier.testTag("parent")) {
         Dialog(onDismissRequest = {}) {
           Box(Modifier.testTag("child"))
@@ -393,8 +396,8 @@ class ComposeUiTest {
 
     assertThat(hierarchy).isEqualTo(
       """
-      |CompositionLocalProvider:
-      |${BLANK}CompositionLocalProvider { test-tag:"parent" }
+      |Box:
+      |${BLANK}Box { test-tag:"parent" }
       |${BLANK}╰─Dialog
       |${BLANK}  ╰─CompositionLocalProvider { DIALOG }
       |${BLANK}    ╰─Box { test-tag:"child" }
@@ -409,6 +412,8 @@ class ComposeUiTest {
     }
 
     composeRule.setContent {
+      currentComposer.collectParameterInformation()
+
       Box(Modifier.testTag("parent")) {
         CustomTestDialog {
           Box(Modifier.testTag("child"))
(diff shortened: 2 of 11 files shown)


buildSrc/src/main/java/Dependencies.kt
```diff
@@ -6,7 +6,8 @@ object Versions {
   val KotlinCompiler = System.getProperty("square.kotlinVersion") ?: "1.9.10"
 
   const val AndroidXTest = "1.5.0"
-  const val Compose = "1.5.3"
+  const val Compose = "1.7.0-beta05"
+  const val ComposeCompiler = "1.5.3"
```
elihart, line 10: I didn't want to attempt to update kotlin at the same time, so I left the compose compiler version

compose-tests/src/androidTest/java/radiography/test/compose/ComposeUiTest.kt
```diff
@@ -393,8 +396,8 @@ class ComposeUiTest {
 
     assertThat(hierarchy).isEqualTo(
       """
-      |CompositionLocalProvider:
-      |${BLANK}CompositionLocalProvider { test-tag:"parent" }
+      |Box:
```
elihart, line 399: I'm not sure where this change comes from exactly, but it seems to be more correct. I am guessing it is a behavior change within compose that provides more accurate naming for nodes some how. I didn't look into it further though

compose-tests/src/androidTest/java/radiography/test/compose/ComposeUiTest.kt
```diff
@@ -595,9 +612,9 @@ class ComposeUiTest {
 
     assertThat(hierarchy).isEqualTo(
       """
-      |CompositionLocalProvider:
-      |${BLANK}CompositionLocalProvider { test-tag:"parent" }
-      |${BLANK}╰─LazyColumn { test-tag:"list", vertical-scroll-axis-range:"ScrollAxisRange(value=0.0, maxValue=0.0)" }
+      |Box:
+      |${BLANK}Box { test-tag:"parent" }
+      |${BLANK}╰─LazyColumn { vertical-scroll-axis-range:"ScrollAxisRange(value=0.0, maxValue=0.0)", test-tag:"list" }
```
elihart, line 617: the order of the attributes changed, so I updated the expected value to match. Perhaps the renderer should sort them by name to keep them consistent?

radiography/src/main/java/radiography/ExperimentalRadiographyComposeApi.kt
```diff
@@ -4,4 +4,5 @@ package radiography
   message = "This API is experimental, may only work with a specific version of Compose, " +
     "and may change or break at any time. Use with caution."
 )
+@Retention(AnnotationRetention.BINARY)
```
elihart, line 7: Lint complained that experimental annotations should have binary retention

radiography/src/main/java/radiography/ViewStateRenderers.kt
```diff
@@ -72,11 +71,10 @@ public object ViewStateRenderers {
 
       // Semantics
       composeView
-        .modifiers
-        .filterIsInstance<SemanticsModifier>()
+        .semanticsConfigurations
```
elihart, line 74: this, and the change in [Semantics.kt](https://github.com/square/radiography/pull/169/files#diff-2b27936776ce7e7111731ed2ede9ca41c137ef09803bad6d1a7a99f9dd8dac7b), are the main changes to support 1.7, where we need to use the semantics nodes instead of the modifiers

radiography/src/main/java/radiography/ScannableView.kt
```diff
@@ -68,9 +70,25 @@ public sealed class ScannableView {
...
         return semanticsNodes.map { it.config }.ifEmpty { null }
+          // Before Compose 1.6, semantics could be held in a SemanticsModifier
           ?: modifiers
             .filterIsInstance<SemanticsModifier>()
             .map { it.semanticsConfiguration }
+            .ifEmpty { null }
+            // After Compose 1.6, semantics are held in nodes. While this usually gets represented
+            // by semanticsNodes, sometimes they instead are only found in SemanticsModifierNode,
+            // which we must then extract.
+          ?: modifiers
```
elihart, line 81: I needed to add this to get a few tests to pass. In some cases, the semanticsNodes are not present for a layout node (and I'm not sure why). However, I noticed that they are instead found within a special type of modifier node that we can leverage.

I would love if someone more familiar with compose internals can validate this approach. This fixes all test cases, but I am not 100% confident that it is the best way to do so. Ideally I think the `computeLayoutInfos` would be able to identify the semanticsNodes for each LayoutNodeInfo instead of needing to do this

radiography/src/main/java/radiography/ScannableView.kt
```diff
@@ -68,9 +70,25 @@ public sealed class ScannableView {
...
+            .filterIsInstance<ModifierNodeElement<*>>()
+            .map { it.create() }
+            .filterIsInstance<SemanticsModifierNode>()
+            .map { semanticsModifierNode ->
+              semanticsModifierNode.run {
+                val semanticsConfiguration = SemanticsConfiguration()
```
elihart, line 87: this creates a new `SemanticsConfiguration()` for each node - I wasn't sure if that was best, or if we should use the same SemanticsConfiguration() instance and apply each semantics node to it. having a list of `SemanticsConfiguration` seems to be consistent with previous behavior though

rjrjr:
Thanks Eli! @pyricau @zach-klippenstein, do you have bandwidth to review?

compose-tests/src/androidTest/java/radiography/test/compose/ComposeUiTest.kt
zach-klippenstein: That seems like a good idea, but we can fix in a follow-up.

radiography/src/main/java/radiography/ScannableView.kt
zach-klippenstein: I'm not sure exactly how we'd want to handle semantics merging here. I think the existing Radiography implementation uses merged config, but since it also gets the configuration per-node, it might not be what "merged config" usually means? It might not matter much - since we _are_ already grouping by node, we can probably just always merge semantics for all modifier nodes on the current node. So we'd probably want to share a `SemanticsConfiguration` object for everything on the same node, and make sure we're calling the modifiers in the right order. We might also need to watch out for `clearAndSetSemantics`, but again I'm not sure we care for Radiography's use case.

zach-klippenstein (approved):
Test failure looks like infra. Left a comment about `SemanticsModifierNode` traversal, but that's something we can tweak as necessary (and should probably add more test coverage for as we do), so no need to block this PR.

PR #194, opened 2026-05-16
Support Compose 1.11 and deterministic subcomposition ordering

zirman:
- Upgrade Compose dependency to 1.11.1.
- Update reflection logic in `CompositionContexts.kt` to support `GapComposer`, `LinkComposer`, and the `RememberObserverHolder` wrapper introduced in newer Compose versions.
- Implement deterministic sorting of subcompositions by child placement order using reflection to access `LayoutNode.getChildren`.
- Update instrumented tests to use `v2` Compose test rules.
- Update expected test output to reflect internal Compose naming changes, such as `ProvideCompositionLocals`.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -20,7 +20,7 @@ android.useAndroidX=true
 systemProp.org.gradle.internal.publish.checksums.insecure=true
 
 GROUP=com.squareup.radiography
-VERSION_NAME=2.9
+VERSION_NAME=2.10-SNAPSHOT
 
 POM_DESCRIPTION=Pretty printing of view hierarchies
 


radiography/src/main/java/radiography/internal/ComposeLayoutInfo.kt
```diff
@@ -34,12 +36,14 @@ import radiography.internal.ComposeLayoutInfo.SubcompositionInfo
...
-      val children: Sequence<ComposeLayoutInfo>,
-      val semanticsNodes: List<SemanticsNode>,
+    val name: String,
+    val callChain: List<CallGroupInfo>,
+    val bounds: IntRect,
+    val modifiers: List<Modifier>,
+    val children: Sequence<ComposeLayoutInfo>,
+    val semanticsNodes: List<SemanticsNode>,
+    // Node reference used to sort Subcompose Layouts deterministically across runs
+    val layoutNode: Any? = null,
```
zach-klippenstein:
```suggestion
  /**
   * @param layoutNode Node reference used to sort Subcompose Layouts deterministically across runs.
   */
  data class LayoutNodeInfo(
    val name: String,
    val callChain: List<CallGroupInfo>,
    val bounds: IntRect,
    val modifiers: List<Modifier>,
    val children: Sequence<ComposeLayoutInfo>,
    val semanticsNodes: List<SemanticsNode>,
    val layoutNode: Any? = null,
```

zach-klippenstein (commented):
Thanks for this! Please write tests for subcomposition ordering, if at all possible.

zirman:
> Thanks for this! Please write tests for subcomposition ordering, if at all possible.

I added a test but it's doesn't prove determinism. We have a large proprietary code base that we snapshot which is my real world test. So far these changes have generated the same order over multiple runs on different environments. I'm going to open a new PR since I broke this one from a force push when trying to sign commits. Thanks.

